PR #102, opened 2014-01-21
Remove uses of AppKit in ShipIt because AppKit isn’t daemon safe

keithduncan:
It unfortunately still links AppKit due to ReactiveCocoa depending on AppKit, but this removes the API uses, `NSWorkspace` and `NSRunningApplication`.

Introduces a new background process to do the process waiting in the user launchd domain, and writes a file when it’s ready. The installer then listens for this file rather than the applications.

---

- [ ] Make installer launched on demand using IPC
- [ ] Make new watcher launched `$MAX` times which is always 'userspace'


joshaber:
RIP the dreams :cry:

PR #138, opened 2014-12-13
Stop checking for updates after downloading one

### LINKED ISSUE #111: Should only download an update once
If the check interval ticks over multiple times, ensure that the same update isn’t downloaded and prepared both times. It would be a waste of bandwidth to download the same update twice.

inket:
Fixes #111 

There's still the issue where a user who almost never restarts the app might get an 'old' update if several updates were released while the app is open. Ideally, we'd compare `pub_date`s to see if the downloaded update should be replaced with a new one.


joshaber:
Superseded by #171.

PR #140, opened 2015-01-10
Let Xcode code sign ShipIt.

joshaber:
Xcode wants to do this and it seems to work so let’s do it what could go wrong gang.


jspahrsummers:
:ok:

jspahrsummers:
Is there any reason to keep `deep-codesign` now?

joshaber:
@keithduncan can answer that better than I.

PR #141, opened 2015-01-14
Separate the update available and install instructions

keithduncan:
Don’t present these in the same section to clarify that the install should be performed after an update has been received.

### DIFF
--- a/README.md
+++ b/README.md
@@ -113,13 +113,18 @@ To know when an update is ready to be installed, you can subscribe to the
 }];
 ```
 
-## Installing Updates on Demand
+## Installing Updates
 
-Once you've been notified of an available update, you may want to present an interface
-informing the user about the update and offers the ability to install and relaunch.
+While downloaded updates are automatically installed when your application
+terminates, if don't want to wait you can terminate manually the app to begin
+the installation process immediately.
+
+Once an [update available notification](#update-available-notifications) has
+been received, you may want to present an interface informing the user about
+the update and offering the ability to install and relaunch.
 
 To install a downloaded update and automatically relaunch afterward, subscribe to
-`relaunchToInstallUpdate` on `SQRLUpdater`:
+the `relaunchToInstallUpdate` signal on `SQRLUpdater`:
 
 ```objc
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -113,12 +113,13 @@ To know when an update is ready to be installed, you can subscribe to the
 }];
 ```
 
-If you've been notified of an available update, and don't want to wait for it to
-be installed automatically, you can terminate the app to begin the installation
-process immediately.
+## Installing Updates on Demand
```
jspahrsummers: How about changing this to “installing updates,” and restoring the bits about termination automatically resulting in update installation?

jspahrsummers:
:+1:

keithduncan:
:book:

README.md
```diff
@@ -113,12 +113,18 @@ To know when an update is ready to be installed, you can subscribe to the
...
 
-If you've been notified of an available update, and don't want to wait for it to
-be installed automatically, you can terminate the app to begin the installation
-process immediately.
+## Installing Updates
 
-If you want to install a downloaded update and automatically relaunch afterward,
-`SQRLUpdater` can do that:
+While downloaded updates are automatically installed when your application
+terminates, if don't want to wait you can terminate manually the app to begin
```
jspahrsummers: “terminate manually the app?” :trollface:

README.md
```diff
@@ -113,12 +113,18 @@ To know when an update is ready to be installed, you can subscribe to the
...
+the installation process immediately.
+
+Once an [update available notification](#update-available-notifications) has
+been received, you may want to present an interface informing the user about
+the update and offering the ability to install and relaunch.
+
+To install a downloaded update and automatically relaunch afterward, subscribe to
```
jspahrsummers: Maybe “to explicitly install?”

keithduncan:
From an API design perspective it would be interesting to make the `relaunchToInstallUpdate` signal a `RACCommand` accepting an update object in `execute:` thus forcing you to have received an update before you could use it. This would prevent the use of this method when there is no update to install.

jspahrsummers:
It would certainly be interesting. :trollface:

PR #149, opened 2015-06-19
Renames

joshaber:
Sometimes, apps change their names. Sometimes. Squirrel should support that. We only change the name if:

1. The user hasn't renamed the existing app.
2. There's no app already named the new name.

Also, it should be noted that I have no idea what I'm doing.

### DIFF
--- a/Squirrel/SQRLInstaller.m
+++ b/Squirrel/SQRLInstaller.m
@@ -433,6 +433,12 @@ NSString * const SQRLInstallerOwnedBundleKey = @"SQRLInstallerOwnedBundle";
 
 		NSString *newAppName = [sourceExecutableName stringByAppendingPathExtension:@"app"];
 		NSURL *newTargetURL = [[targetURL URLByDeletingLastPathComponent] URLByAppendingPathComponent:newAppName isDirectory:YES];
+		NSBundle *bundleWithNewName = [NSBundle bundleWithURL:newTargetURL];
+		// If there's already an app at that location and its bundle ID doesn't
+		// match ours, leave it alone.
+		if (bundleWithNewName != nil && ![bundleWithNewName.bundleIdentifier isEqual:sourceBundle.bundleIdentifier]) {
+			return [RACSignal return:targetURL];
+		}
 
 		if (rename(targetURL.path.fileSystemRepresentation, newTargetURL.path.fileSystemRepresentation) == 0) {
 			return [RACSignal return:newTargetURL];
--- a/Squirrel/SQRLInstaller.m
+++ b/Squirrel/SQRLInstaller.m
@@ -417,6 +417,21 @@ NSString * const SQRLInstallerOwnedBundleKey = @"SQRLInstallerOwnedBundle";
 	NSParameterAssert(targetURL != nil);
 	NSParameterAssert(sourceURL != nil);
 
+	NSBundle *targetBundle = [NSBundle bundleWithURL:targetURL];
+	NSString *targetExecutableName = [targetBundle objectForInfoDictionaryKey:(id)kCFBundleExecutableKey];
+
+	NSBundle *sourceBundle = [NSBundle bundleWithURL:sourceURL];
+	NSString *sourceExecutableName = [sourceBundle objectForInfoDictionaryKey:(id)kCFBundleExecutableKey];
+	// Only rename if the existing app is named after its executable.
+	if (targetExecutableName != nil && ![targetExecutableName isEqual:sourceExecutableName]) {
+		NSString *targetAppName = [targetExecutableName stringByAppendingPathExtension:@"app"];
+		if ([targetAppName isEqual:targetURL.lastPathComponent]) {
+			NSURL *oldTargetURL = targetURL;
+			targetURL = [[targetURL URLByDeletingLastPathComponent] URLByAppendingPathComponent:targetAppName isDirectory:YES];
+			rename(oldTargetURL.path.fileSystemRepresentation, targetURL.path.fileSystemRepresentation);
+		}
+	}
+
 	return [[[[RACSignal
 		defer:^{
 			// rename() is atomic, NSFileManager sucks.
--- a/Squirrel/SQRLInstaller.m
+++ b/Squirrel/SQRLInstaller.m
@@ -271,12 +271,12 @@ NSString * const SQRLInstallerOwnedBundleKey = @"SQRLInstallerOwnedBundle";
 			return [[[[self
 				acquireTargetBundleURLForRequest:request]
 				then:^{
-					if (!request.allowRename) return [RACSignal return:request.targetBundleURL];
+					if (!request.useUpdateBundleName) return [RACSignal return:request.targetBundleURL];
 
 					return [self renamedTargetIfNeededWithTargetURL:request.targetBundleURL sourceURL:updateBundleURL];
 				}]
 				flattenMap:^(NSURL *newTargetURL) {
-					SQRLShipItRequest *updatedRequest = [[SQRLShipItRequest alloc] initWithUpdateBundleURL:request.updateBundleURL targetBundleURL:newTargetURL bundleIdentifier:request.bundleIdentifier launchAfterInstallation:request.launchAfterInstallation allowRename:request.allowRename];
+					SQRLShipItRequest *updatedRequest = [[SQRLShipItRequest alloc] initWithUpdateBundleURL:request.updateBundleURL targetBundleURL:newTargetURL bundleIdentifier:request.bundleIdentifier launchAfterInstallation:request.launchAfterInstallation useUpdateBundleName:request.useUpdateBundleName];
 					return [[[[[[[self
 						installItemToURL:request.targetBundleURL fromURL:updateBundleURL]
 						then:^{
--- a/Squirrel/SQRLInstaller.m
+++ b/Squirrel/SQRLInstaller.m
@@ -413,27 +413,46 @@ NSString * const SQRLInstallerOwnedBundleKey = @"SQRLInstallerOwnedBundle";
 
 #pragma mark Installation
 
+- (RACSignal *)renameIfNeededWithTargetURL:(NSURL *)targetURL sourceURL:(NSURL *)sourceURL {
+	return [RACSignal defer:^{
+		NSBundle *targetBundle = [NSBundle bundleWithURL:targetURL];
+		NSString *targetExecutableName = [targetBundle objectForInfoDictionaryKey:(id)kCFBundleExecutableKey];
+
+		NSBundle *sourceBundle = [NSBundle bundleWithURL:sourceURL];
+		NSString *sourceExecutableName = [sourceBundle objectForInfoDictionaryKey:(id)kCFBundleExecutableKey];
+		// Only rename if the existing app is named after its executable.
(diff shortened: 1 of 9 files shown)


Squirrel/SQRLInstaller.m
```diff
@@ -413,6 +428,28 @@ - (RACSignal *)verifyBundleAtURL:(NSURL *)bundleURL usingSignature:(SQRLCodeSign
 
 #pragma mark Installation
 
+- (RACSignal *)renameTargetIfNeededWithTargetURL:(NSURL *)targetURL sourceURL:(NSURL *)sourceURL {
```
mdiep: Maybe call this `renamed` target to make it sound less like it's renaming the target on disk?

Also, this method could really use documentation.

Squirrel/SQRLUpdater.m
```diff
@@ -528,7 +528,9 @@ - (RACSignal *)prepareUpdateForInstallation:(SQRLDownloadedUpdate *)update {
 	return [[[[RACSignal
 		defer:^{
 			NSRunningApplication *currentApplication = NSRunningApplication.currentApplication;
-			SQRLShipItRequest *request = [[SQRLShipItRequest alloc] initWithUpdateBundleURL:update.bundle.bundleURL targetBundleURL:currentApplication.bundleURL bundleIdentifier:currentApplication.bundleIdentifier launchAfterInstallation:NO];
+			NSBundle *appBundle = [NSBundle bundleWithURL:currentApplication.bundleURL];
+			BOOL allowRename = [appBundle.sqrl_executableName isEqual:currentApplication.bundleURL.lastPathComponent.stringByDeletingPathExtension];
```
mdiep: Is the same logic as above?

```objc
+		// If they're already the same then we're good.
+		if ([targetExecutableName isEqual:sourceExecutableName]) {
+			return [RACSignal return:targetURL];
+		
```

joshaber: Not quite. This is checking to see if the existing app bundle's name matches its `CFBundleExecutable` name. I.e., has the user renamed the app?

mdiep: Ah, right. :+1: Can you add a comment to that effect?

Squirrel/SQRLShipItRequest.h
```diff
@@ -73,10 +73,12 @@ extern NSString * const SQRLShipItRequestPropertyErrorKey;
 //                           installing. Can be nil.
 // launchAfterInstallation - Whether the updated application should be launched
 //                           after installation.
+// allowRename             - Should the target be renamed if it doesn't match
+//                           the update?
 //
 // Returns a request which can be written to disk for ShipIt to read and
 // perform.
-- (instancetype)initWithUpdateBundleURL:(NSURL *)updateBundleURL targetBundleURL:(NSURL *)targetBundleURL bundleIdentifier:(NSString *)bundleIdentifier launchAfterInstallation:(BOOL)launchAfterInstallation;
+- (instancetype)initWithUpdateBundleURL:(NSURL *)updateBundleURL targetBundleURL:(NSURL *)targetBundleURL bundleIdentifier:(NSString *)bundleIdentifier launchAfterInstallation:(BOOL)launchAfterInstallation allowRename:(BOOL)allowRename;
```
keithduncan: I think I’d like to call this `useUpdateBundleName` so that it’s clearer what the final name will be, `rename` here could imply using or not using `rename(2)`.

Alternatively, we could be even more descriptive and make it a `targetPathComponent: NSString?` parameter. This would indicate the `lastPathComponent` on top of the `targetBundleURL.stringByRemovingLastPathComponent` that we’re requesting the update service to put the update into.

Squirrel/SQRLInstaller.m
```diff
@@ -413,6 +428,36 @@ - (RACSignal *)verifyBundleAtURL:(NSURL *)bundleURL usingSignature:(SQRLCodeSign
...
+		// If they're already the same then we're good.
+		if ([targetExecutableName isEqual:sourceExecutableName]) {
+			return [RACSignal return:targetURL];
+		}
+
+		NSString *newAppName = [sourceExecutableName stringByAppendingPathExtension:@"app"];
+		NSURL *newTargetURL = [targetURL.URLByDeletingLastPathComponent URLByAppendingPathComponent:newAppName];
+
+		// If there's already something there then don't rename to it.
+		if ([NSFileManager.defaultManager fileExistsAtPath:newTargetURL.path]) {
```
keithduncan: This is a minor ToC-ToU race, to prevent it I’d like to flip the order of installation and rename.

In my head, I’ve been modelling the file system as a datastore whose entries are protected by code sign requirements; entries that don’t exist have no protecting requirement (only BSD and ACL permissions).

If we can safely (i.e. don't rename from the current entry to the new one if the new one already exists) the existing bundle to the requested name, the install is then from the update bundle (whose name is irrelevant at that point) to the file system entry of the target bundle, as before.

joshaber: Yeah so I started down that route originally. The problem we run into is that the `ShipIt` job has a hard-coded path that includes the original bundle path: https://github.com/Squirrel/Squirrel.Mac/blob/9e040347f5766ece29044a29b6286c11af202370/Squirrel/SQRLShipItLauncher.m#L54

So I *think* we could end up in a situation where:

1. We rename.
2. Something fails.
3. `launchd` can't find `ShipIt` to retry the update installation or recover since the bundle's been renamed.

Now, we have the same problem if we rename after installation, except that we're less likely to run into a problem at that point (though still possible).

Or at least that's my current understanding.

keithduncan: > launchd can't find ShipIt to retry the update installation or recover since the bundle's been renamed.

That’s a great point, I hadn’t considered that :cry:

> But for that matter, I'm not sure how swapping the order solves the potential ToC-ToU race. We still have to check for existence before renaming, regardless of when the rename happens.

It would be handled by instead of doing a `rename(2)` into the unoccupied file system location, doing a move-if-not-exist (I actually couldn’t find a way to do this atomically since `link(2)` doesn’t work with directories). This moves the codesign reservation from the old name to the new name and the install continues as normal. If the move failed we’d fall back to updating in place.

My concern is renaming could be exploited to replace a different app, which the update is named for but doesn’t meet the codesign requirements of.

joshaber:
Alright @keithduncan, see how that strikes you.

We still have the `launchd` problem, but I wanted to make sure this addressed your other concern first.

Squirrel/SQRLInstaller.m
```diff
@@ -413,6 +435,36 @@ - (RACSignal *)verifyBundleAtURL:(NSURL *)bundleURL usingSignature:(SQRLCodeSign
...
+		NSBundle *sourceBundle = [NSBundle bundleWithURL:sourceURL];
+		NSString *targetExecutableName = targetURL.lastPathComponent.stringByDeletingPathExtension;
+		NSString *sourceExecutableName = sourceBundle.sqrl_executableName;
+
+		// If they're already the same then we're good.
+		if ([targetExecutableName isEqual:sourceExecutableName]) {
+			return [RACSignal return:targetURL];
+		}
+
+		NSString *newAppName = [sourceExecutableName stringByAppendingPathExtension:@"app"];
```
keithduncan: Can we use `sourceURL.lastPathComponent` here?

Squirrel/SQRLInstaller.m
```diff
@@ -413,6 +435,36 @@ - (RACSignal *)verifyBundleAtURL:(NSURL *)bundleURL usingSignature:(SQRLCodeSign
...
+/// Returns a signal which will send the URL for the renamed target. If a rename
+/// isn't needed then it will send `targetURL`.
+- (RACSignal *)renamedTargetIfNeededWithTargetURL:(NSURL *)targetURL sourceURL:(NSURL *)sourceURL {
+	return [RACSignal defer:^{
+		NSBundle *sourceBundle = [NSBundle bundleWithURL:sourceURL];
+		NSString *targetExecutableName = targetURL.lastPathComponent.stringByDeletingPathExtension;
+		NSString *sourceExecutableName = sourceBundle.sqrl_executableName;
+
+		// If they're already the same then we're good.
+		if ([targetExecutableName isEqual:sourceExecutableName]) {
```
keithduncan: Similarly should this compare last components rather than executable names?

keithduncan:
I’d like to get another :+1: from @jspahrsummers (if still interested in maintaining Squirrel.Mac).

What do you think we should do about the launchd concern?

Squirrel/SQRLInstaller.m
joshaber: I think the bundle's executable name is more canonical than what the zipped thing is.

Squirrel/SQRLInstaller.m
joshaber: Same as above. The plist is more canonical than whatever the name of the zipped thing is, I think.

joshaber:
>What do you think we should do about the launchd concern?

Yeah so this is interesting. It *seems* like Squirrel has always had this problem.

Before:

1. Start the update.
1. "Acquire" the bundle to update, which really means move it.
1. :boom: `ShipIt` is no longer where we registered it.

Now:

1. Start the update.
1. Rename the thing if needed, or "acquire" the bundle.
1. :boom: `ShipIt` is no longer where we registered it.

Both before and now, after the first step in the update process there's no going back.

We could try re-registering the job with the new path, but it's not clear to me if we can since it'd be a different bundle than the original one.

But at any rate, it _seems_ like a pre-existing condition so I'd prefer to get this merged and address it later.

Squirrel/SQRLInstaller.m
keithduncan: @mdiep could you tie break on this one

keithduncan:
> Yeah so this is interesting. It seems like Squirrel has always had this problem.

Weird, you’re totally right :confounded:

I’ll wait to see what @mdiep says about which name to prefer then we can :shipit:

Squirrel/SQRLInstaller.m
mdiep: I don't know why these would ever be different at this point. It seems a little weird to me that this is even a key in the `Info.plist`. But given that it's there, I agree that it seems more canonical than the URL.

keithduncan:
shia_labeouf_delivers_the_most_intense_motivational_speech_of_all_time

jspahrsummers:
IIRC launchd copies the ShipIt executable when the job is submitted.

keithduncan:
```
$ launchctl list com.github.GitHub.ShipIt
{
	"StandardOutPath" = "/Users/keith/Library/Application Support/com.github.GitHub.ShipIt/ShipIt_stdout.log";
	"LimitLoadToSessionType" = "Aqua";
	"StandardErrorPath" = "/Users/keith/Library/Application Support/com.github.GitHub.ShipIt/ShipIt_stderr.log";
	"MachServices" = {
		"com.github.GitHub.ShipIt" = mach-port-object;
	};
	"Label" = "com.github.GitHub.ShipIt";
	"TimeOut" = 30;
	"OnDemand" = true;
	"LastExitStatus" = 0;
	"Program" = "/Users/keith/Downloads/GitHub.app/Contents/Frameworks/Squirrel.framework/Resources/ShipIt";
	"ProgramArguments" = (
		"/Users/keith/Downloads/GitHub.app/Contents/Frameworks/Squirrel.framework/Resources/ShipIt";
		"com.github.GitHub.ShipIt";
	);
};
```

Nope :confused:

PR #152, opened 2015-09-21
Fix pruning

joshaber:
Because we replay, we don’t ever prune after the first subscription. We just keep replaying the previous results. So we end up accumulating a ton of updates.

Fixes https://github.com/atom/atom/issues/7061.

### DIFF
--- a/Squirrel/SQRLUpdater.m
+++ b/Squirrel/SQRLUpdater.m
@@ -49,16 +49,6 @@ static NSString * const SQRLUpdaterUniqueTemporaryDirectoryPrefix = @"update.";
 // Sends completed or error.
 @property (nonatomic, strong, readonly) RACSignal *shipItLauncher;
 
-// Lazily removes outdated temporary directories (used for previous updates)
-// upon first subscription.
-//
-// Pruning directories while an update is pending or in progress will result in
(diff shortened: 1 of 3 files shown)


Squirrel/SQRLUpdater.m
```diff
@@ -169,7 +169,7 @@ - (id)initWithUpdateRequest:(NSURLRequest *)updateRequest {
 	BOOL updatesDisabled = (getenv("DISABLE_UPDATE_CHECK") != NULL);
 	@weakify(self);
 
-	_prunedUpdateDirectories = [[[[RACSignal
+	_prunedUpdateDirectories = [[[RACSignal
```
mdiep: Now that this isn't idempotent, can we make it a method instead? And maybe name it `pruneUpdateDirectories` so that it's clear that it's doing the pruning?

mdiep:
Good catch. :metal:

kevinsawicki:
:blue_heart: thanks so much for figuring this one out :+1:

Squirrel/SQRLUpdater.m
joshaber: Good shout :+1:

PR #171, opened 2016-06-01
Conditionally download updates.

### LINKED ISSUE #111: Should only download an update once

joshaber:
Fixes #111

### DIFF
--- a/Squirrel/SQRLUpdater.m
+++ b/Squirrel/SQRLUpdater.m
@@ -364,8 +364,6 @@ static NSString * const SQRLUpdaterUniqueTemporaryDirectoryPrefix = @"update.";
 				}]
 				flatten]
 				flattenMap:^(NSData *data) {
-					if (data.length == 0) return [RACSignal empty];
-
 					NSURL *zipOutputURL = [downloadDirectory URLByAppendingPathComponent:zipDownloadURL.lastPathComponent];
 
 					NSError *error = nil; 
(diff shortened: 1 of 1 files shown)


joshaber:
This assumes we can trust servers to return and respect etags. Is that a reasonable assumption these days?

joshaber:
Looks like S3 respects etags which I gotta think is where most updates live.

Squirrel/SQRLUpdater.m
```diff
@@ -40,6 +40,10 @@ @interface SQRLUpdater ()
 
 @property (atomic, readwrite) SQRLUpdaterState state;
 
+/// The etag of the currently downloaded update, nil if no update has been
+/// downloaded.
+@property (atomic, copy) NSString *etag;
```
keithduncan: Does this need to be atomic, inside the scope of the `RACCommand` shouldn’t access be serialised?

keithduncan:
I don’t think we lose anything with the `atomic` and I don’t want to split hairs, :+1: to merge this

Squirrel/SQRLUpdater.m
joshaber: LOL I honestly couldn't recall anymore, so I went with safe over sorry.

PR #172, opened 2016-06-02
Truncate logs

### LINKED ISSUE #170: Truncate logs
I've seen a few reports of log files that were 1+ gigs on disk. We should really truncate them.

joshaber:
Fixes #170

### DIFF
--- a/Squirrel/SQRLUpdater.m
+++ b/Squirrel/SQRLUpdater.m
@@ -536,7 +536,6 @@ static NSString * const SQRLUpdaterUniqueTemporaryDirectoryPrefix = @"update.";
 		[handle seekToFileOffset:fileLength - MAX_LENGTH];
 		NSData *mostRecentData = [handle readDataToEndOfFile];
 		[handle truncateFileAtOffset:0];
-		[handle seekToFileOffset:0];
 		[handle writeData:mostRecentData];
 
 		return [RACSignal empty];
--- a/Squirrel/SQRLUpdater.m
+++ b/Squirrel/SQRLUpdater.m
@@ -524,7 +524,7 @@ static NSString * const SQRLUpdaterUniqueTemporaryDirectoryPrefix = @"update.";
 // Like truncation, but backwards.
 - (RACSignal *)backwardTruncateFile:(NSURL *)fileURL {
 	return [RACSignal defer:^{
-		static const NSInteger MAX_LENGTH = 1024 * 8;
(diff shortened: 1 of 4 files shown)


Squirrel/SQRLUpdater.m
```diff
@@ -516,6 +520,39 @@ - (RACSignal *)pruneUpdateDirectories {
...
+		static const NSInteger MAX_LENGTH = 1024 * 8;
+
+		NSError *error;
+		NSFileHandle *handle = [NSFileHandle fileHandleForWritingToURL:fileURL error:&error];
+		if (handle == nil) return [RACSignal error:error];
+
+		unsigned long long fileLength = [handle seekToEndOfFile];
+		if (fileLength <= MAX_LENGTH) return [RACSignal empty];
+
+		[handle seekToFileOffset:fileLength - MAX_LENGTH];
```
joshaber: I don't love this, but I also couldn't find a better way to truncate from the start of a file. Since log messages are appended, we want to keep the most recent messages, i.e., the messages at the end.

Squirrel/SQRLUpdater.m
```diff
@@ -516,6 +520,39 @@ - (RACSignal *)pruneUpdateDirectories {
 		setNameWithFormat:@"%@ -prunedUpdateDirectories", self];
 }
 
+
+// Like truncation, but backwards.
+- (RACSignal *)backwardTruncateFile:(NSURL *)fileURL {
+	return [RACSignal defer:^{
+		static const NSInteger MAX_LENGTH = 1024 * 8;
```
keithduncan: 8k seems like a small max size, could we bump this to 8m?

Squirrel/SQRLUpdater.m
keithduncan: This doesn’t truncate in a line aware fashion which is a bummer, what do you think about taking the last N lines instead?

joshaber: I didn't figure it was worth the hassle. If we need to be line-aware, we either need to read the whole file (bummer if it's already big) or get a lot more complex. I'm not sure it's worth it.

Squirrel/SQRLUpdater.m
joshaber: 👍

joshaber:
🍯

Squirrel/SQRLUpdater.m
keithduncan: 👌 lets roll with the minimum size

PR #175, opened 2016-06-08
Don't prune if we've already downloaded updates.

### LINKED ISSUE #174: Downloaded updates are pruned by subsequent update checks
This was introduced by https://github.com/Squirrel/Squirrel.Mac/pull/171 :see_no_evil::

1. Check for updates.
1. Prune existing downloads.
1. Download an update.
1. Check for updates.
1. Prune existing downloads.
1. Conditional `GET` means we don't re-download.
1. Now we don't have an update downloaded anymore.

joshaber:
Fixes #174


keithduncan:
This looks fine to me, we have other signals in the same chain that check and mutate the state field :+1:

PR #176, opened 2016-06-14
Don't error if our conditional GET fails.

joshaber:
This was caused by #175 and #171.

If we tried to download an update and our conditional GET says we already have it, we’d throw an error later when we tried to find the app bundle that we didn't download: https://github.com/Squirrel/Squirrel.Mac/blob/9ebfd397fd1485d3eb02e7cb3d687f27c995e3b4/Squirrel/SQRLUpdater.m#L460

That meant we wouldn’t end up in the `SQRLUpdaterStateAwaitingRelaunch` state, and so we’d prune the already downloaded update, and still experience #174.

### DIFF
--- a/Squirrel/SQRLUpdater.m
+++ b/Squirrel/SQRLUpdater.m
@@ -304,16 +304,27 @@ static NSString * const SQRLUpdaterUniqueTemporaryDirectoryPrefix = @"update.";
 	return [[[self
 		uniqueTemporaryDirectoryForUpdate]
 		flattenMap:^(NSURL *downloadDirectory) {
+			void (^cleanUp)(void) = ^{
+				NSError *error;
+				if (![NSFileManager.defaultManager removeItemAtURL:downloadDirectory error:&error]) {
+					NSLog(@"Error removing temporary download directory at %@: %@", downloadDirectory, error.sqrl_verboseDescription);
+				}
+			};
+
 			return [[[self
 				downloadBundleForUpdate:update intoDirectory:downloadDirectory]
 				flattenMap:^(NSBundle *updateBundle) {
+					// If the bundle is nil it means our conditional GET told us
+					// we already downloaded the update. So just clean up.
+					if (updateBundle == nil) {
+						cleanUp();
+						return [RACSignal empty];
+					}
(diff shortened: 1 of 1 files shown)


joshaber:
I shoehorned this into an app and it actually does what it should :tada:

Squirrel/SQRLUpdater.m
```diff
@@ -304,16 +304,27 @@ - (RACSignal *)downloadAndPrepareUpdate:(SQRLUpdate *)update {
...
 		flattenMap:^(NSURL *downloadDirectory) {
+			void (^cleanUp)(void) = ^{
+				NSError *error;
+				if (![NSFileManager.defaultManager removeItemAtURL:downloadDirectory error:&error]) {
+					NSLog(@"Error removing temporary download directory at %@: %@", downloadDirectory, error.sqrl_verboseDescription);
+				}
+			};
+
 			return [[[self
 				downloadBundleForUpdate:update intoDirectory:downloadDirectory]
```
keithduncan: Is there any way we could have the download signal return the already downloaded URL from the cache to save these if nil branches?

joshaber: What we really need is a monad transformer for RAC.

But anyways, I like that idea, but essentially all the branches are avoiding doing things with the _zip_, which we understandably don't keep around after unzipping.

keithduncan: > What we really need is a monad transformer for RAC.

Tell me more :thinking_face:

> But anyways, I like that idea, but essentially all the branches are avoiding doing things with the zip

Hmm that’s a good point, can we structure it so that we return a signal from inside the conditional get that is flattened, then we can branch on whether we have a zip to decompress or can fulfill the unarchived update from the cache?

joshaber: I reorganized so we can avoid all the annoying `nil` checks. We're still providing a `nil` bundle because we want to distinguish between the case where we just downloaded the update and want to verify it, and the case where we already have it and just need to clean up the temp directory.

keithduncan:
:+1: this looks good to me

PR #183, opened 2016-09-29
[WIP] Try to unquarantine the host app before checking for updates.

### LINKED ISSUE #182: Gatekeeper Path Randomization makes life hard
1. Use macOS Sierra.
2. Download an app to `~/Downloads`.
3. Launch app.
4. Update found.
5. Prompted to auth update helper (??)
6. Update can't be installed.

This is because of Gatekeeper's path randomization, new in Sierra. Once you do something that deactivates path randomization (e.g., move the app), updates are fine.

This sucks. It's not terribly clear how we can deal with it. Try to detect it and tell users? Maybe we can explicitly de-quarantine the running app from ShipIt?

joshaber:
Fixes #182

### DIFF
--- a/Squirrel/SQRLUpdater.m
+++ b/Squirrel/SQRLUpdater.m
@@ -6,6 +6,7 @@
 //  Copyright (c) 2013 GitHub. All rights reserved.
 //
 
+#import <sys/xattr.h>
 #import "SQRLUpdater.h"
 #import "NSBundle+SQRLVersionExtensions.h"
 #import "NSError+SQRLVerbosityExtensions.h"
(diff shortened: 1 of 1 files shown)


joshaber:
>I'm surprised an app is allowed to self remove the quarantine but though, that would appear to defeat the protection aspect too

Yeah I'm curious to see if this actually works once I figure out how to test it.

keithduncan:
> I've confirmed that removing com.apple.quarantine from the app fixes this

Ah I misinterpreted this as you having tested it. Right on, I suspect this will in fact not work then :wink:

joshaber:
Yeah, I mean, thinking about this, surely `NSRunningApplication.currentApplication.bundleURL` would be the randomized path, in which case `removexattr` on it is just lulz.

keithduncan:
> Yeah, I mean, thinking about this, surely NSRunningApplication.currentApplication.bundleURL would be the randomized path, in which case removexattr on it is just a joke.

True that, does path randomisation kick in even when you’re sandboxed?

/cc @paulcbetts for details on the magic sandbox update entitlement

joshaber:
Superseded by #186

PR #186, opened 2016-10-14
Error if there's an update available but the app's on a read-only volume

joshaber:
Addresses #182

I _think_ it makes sense to only error once we know that it matters, i.e., that there's an update available. Otherwise it'd just be an annoying error when the user first downloads the app.

The read-only volume check is taken from Sparkle: https://github.com/sparkle-project/Sparkle/blob/3a5c620b60f483b71f8c28573ac29bf85fda6193/Sparkle/SUHost.m#L178-L183

### DIFF
--- a/Squirrel/SQRLUpdater.m
+++ b/Squirrel/SQRLUpdater.m
@@ -33,6 +33,9 @@ const NSInteger SQRLUpdaterErrorInvalidServerResponse = 5;
 const NSInteger SQRLUpdaterErrorInvalidJSON = 6;
 const NSInteger SQRLUpdaterErrorInvalidServerBody = 7;
 
+/// The application's being run on a read-only volume.
+const NSInteger SQRLUpdaterErrorReadOnlyVolume = 7;
+
 // The prefix used when creating temporary directories for updates. This will be
 // followed by a random string of characters.
 static NSString * const SQRLUpdaterUniqueTemporaryDirectoryPrefix = @"update.";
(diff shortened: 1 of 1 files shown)


Squirrel/SQRLUpdater.m
```diff
@@ -32,6 +33,9 @@
 const NSInteger SQRLUpdaterErrorInvalidJSON = 6;
 const NSInteger SQRLUpdaterErrorInvalidServerBody = 7;
 
+/// The application's being run on a read-only volume.
+const NSInteger SQRLUpdaterErrorReadOnlyVolume = 7;
```
keithduncan: Should this be `8` to distinguish from `SQRLUpdaterErrorInvalidServerBody`

Squirrel/SQRLUpdater.m
```diff
@@ -194,6 +198,16 @@ - (id)initWithUpdateRequest:(NSURLRequest *)updateRequest {
 					if (httpResponse.statusCode == 204 /* No Content */) {
 						return [RACSignal empty];
 					}
+
+					BOOL readOnlyVolume = [self isRunningOnReadOnlyVolume];
+					if (readOnlyVolume) {
+						NSDictionary *errorInfo = @{
+						NSLocalizedDescriptionKey: NSLocalizedString(@"Read-only volume", nil),
```
keithduncan: Could you make this `@"Cannot update while running on a read-only volume"`

Squirrel/SQRLUpdater.m
```diff
@@ -492,6 +506,14 @@ - (RACSignal *)shipItStateURL {
 		setNameWithFormat:@"%@ -shipItStateURL", self];
 }
 
+/// Is the host app running on a read-only volume?
+- (BOOL)isRunningOnReadOnlyVolume {
+	struct statfs statfsInfo;
+	NSURL *bundleURL = NSRunningApplication.currentApplication.bundleURL;
+	statfs(bundleURL.fileSystemRepresentation, &statfsInfo);
```
keithduncan: This syscall should be error handled, should we default to `true` in the error case?

Squirrel/SQRLUpdater.m
joshaber: Laugh on line!!

joshaber:
🗝

keithduncan:
urban_cookie_collective

PR #197, opened 2017-01-03
fix SMJobSubmit fail on some osx

sunuslee:
SMJobSubmit will fail on some osx(tested in 10.11).
when this happened, the job will be shown in the launchctl list
but will never be executed (ShipIt not running) and the process
will be terminated by later [NSApp terminate:self] as expected.
so we manually launchctl start the ShipIt job.

Signed-off-by: sunuslee <sunuslee@gmail.com>

### DIFF
--- a/Squirrel/SQRLShipItLauncher.m
+++ b/Squirrel/SQRLShipItLauncher.m
@@ -117,6 +117,15 @@ const NSInteger SQRLShipItLauncherErrorCouldNotStartService = 1;
 		setNameWithFormat:@"+shipItAuthorization"];
 }
 
++ (BOOL)isShipItRunning {
+	NSTask *checkShipItRunningTask = [[NSTask alloc] init];
+	checkShipItRunningTask.launchPath=@"/usr/bin/pgrep";
+	checkShipItRunningTask.arguments = @[@"ShipIt"];
+	[checkShipItRunningTask launch];
+	[checkShipItRunningTask waitUntilExit];
+	return checkShipItRunningTask.terminationStatus == 0;
+}
+
 + (RACSignal *)launchPrivileged:(BOOL)privileged {
 	return [[[RACSignal
 		zip:@[
@@ -141,10 +150,30 @@ const NSInteger SQRLShipItLauncherErrorCouldNotStartService = 1;
 				return [RACSignal error:CFBridgingRelease(cfError)];
 			}
 
+            // wait for ShipIt to start.
+            sleep(1);
+
+            if (![self isShipItRunning]) {
+				// ShipIt not running, possible something wrong with SumJobSubmit
+				// running manually
+				NSLog(@"ShipIt is not running. try to launch");
+				NSTask *launchctlTask = [[NSTask alloc] init];
+				launchctlTask.launchPath = @"/bin/launchctl";
+				launchctlTask.arguments = @[@"start", self.shipItJobLabel];
+				[launchctlTask launch];
+				[launchctlTask waitUntilExit];
+				if (launchctlTask.terminationStatus == 0) {
+					NSLog(@"ShipIt is running");
+				} else {
+					NSLog(@"ShipIt is failed to launch %d", launchctlTask.terminationStatus);
+				}
+			}
+
 			return [RACSignal empty];
 		}]
 		flatten]
 		setNameWithFormat:@"+launchPrivileged: %i", (int)privileged];
 }
+
 
 @end


joshaber:
I'm pretty conservative about making changes to Squirrel.Mac, given:

1. The original maintainers are gone.
2. A lot of projects depend on it.

So to consider this PR, I'll need a more thorough description of the bug it's fixing.

Squirrel/SQRLShipItLauncher.m
```diff
@@ -141,10 +150,30 @@ + (RACSignal *)launchPrivileged:(BOOL)privileged {
 				return [RACSignal error:CFBridgingRelease(cfError)];
 			}
 
+            // wait for ShipIt to start.
+            sleep(1);
```
joshaber: This seems to open us up to a race condition.

Squirrel/SQRLShipItLauncher.m
```diff
@@ -117,6 +117,15 @@ + (RACSignal *)shipItAuthorization {
 		setNameWithFormat:@"+shipItAuthorization"];
 }
 
++ (BOOL)isShipItRunning {
+	NSTask *checkShipItRunningTask = [[NSTask alloc] init];
+	checkShipItRunningTask.launchPath=@"/usr/bin/pgrep";
```
joshaber: Can we use `NSRunningApplication` or `NSWorkspace` or something higher-level than shelling out to `pgrep`?

Squirrel/SQRLShipItLauncher.m
```diff
@@ -117,6 +117,15 @@ + (RACSignal *)shipItAuthorization {
...
+	checkShipItRunningTask.arguments = @[@"ShipIt"];
+	[checkShipItRunningTask launch];
+	[checkShipItRunningTask waitUntilExit];
```
joshaber: This shouldn't be blocking.

Squirrel/SQRLShipItLauncher.m
```diff
@@ -141,10 +150,30 @@ + (RACSignal *)launchPrivileged:(BOOL)privileged {
 				return [RACSignal error:CFBridgingRelease(cfError)];
 			}
 
+            // wait for ShipIt to start.
+            sleep(1);
+
+            if (![self isShipItRunning]) {
+				// ShipIt not running, possible something wrong with SumJobSubmit
```
joshaber: Typo: `SMJobSubmit`

Squirrel/SQRLShipItLauncher.m
sunuslee: hi, i am aware of NSRunningApplication and NSWorkspace. but **ShipIt** is a command line tool. it can not be found by `[[NSWorkspace sharedWorkspace]runningApplications];` and it does not have a valid bundle id. so, we have to search by it's name.
correct me if i am wrong:)

Squirrel/SQRLShipItLauncher.m
sunuslee: during our own test, there might be some time before launchctl submit bring up ShipIt binary. so we just wait , for now. 
otherwise, the check for shipIt is running might get the wrong result. but the update will still work. because what we did in the code is run another shipit.

Squirrel/SQRLShipItLauncher.m
sunuslee: thx man

sunuslee:
@joshaber when this bug happened , if the user finished downloading the new update file, the app will quit and won't launch again. because it's ShipIt's responsibility to bring up the app, but ShipIt can not start running in the first place, because SMJobSumit failed.

joshaber:
>but ShipIt can not start running in the first place, because SMJobSumit failed

But why does it fail?

sunuslee:
@joshaber 
i wish i knew the answer. but it seems like an apple's bug to me.
when you run
`launchctl submit -l test -p /usr/bin/touch  -- /usr/bin/touch  /var/tmp/test`
it should be a file in /var/tmp/test and it will show you this job 

```
$ launchctl list test

{
	"LimitLoadToSessionType" = "Aqua";
	"Label" = "test";
	"TimeOut" = 30;
	"OnDemand" = false;
	"LastExitStatus" = 0;
	"Program" = "/usr/bin/touch";
	"ProgramArguments" = (
		"/usr/bin/touch";
		"/var/tmp/test";
	);
};
```

but when this bug happened, this simple command will only show you the job in launchctl list , and not executing at all.
i wish i knew what's going on inside the system:(

sunuslee:
@joshaber can you offer a better solution?

joshaber:
I can't since I don't fully understand the bug or how it happens. But I'm hesitant to fix an apparent race condition by introducing another potential race condition.

joshaber:
Closing due to inactivity.

PR #206, opened 2017-07-17
Update to Xcode 8

joshaber:
Update the project and dependencies to build with Xcode 8.


joshaber:
It sucks that CI is still broken, but at least we're building with Xcode 8 now.

I suspect the CI script will need to be torn down and rebuilt around whatever it is folks use these days.

PR #207, opened 2017-07-17
Fix installing updates which need escalated privileges

### LINKED ISSUE #131: Failure to install updates that require administrator privileges
Even though we prompt for a password, it apparently still fails. See https://github.com/atom/atom/issues/2860#issuecomment-59517219 and later comments for some example logs.

### LINKED ISSUE #157: Non-admin that owns an app in /Applications can’t update it
I’m a non-admin user (for security purposes), but I know that Atom doesn’t update itself properly when owned by another user, so I made a concession and allowed it to be owned by my non-admin user. I still stuck it in /Applications, which my non-admin user doesn’t have write access to.

Atom still can’t update itself. I get “Restart and Install Update” in the Atom menu, but when I hit it, Atom restarts without having installed an update.

~Library/Application Support/com.github.atom.ShipIt/ShipIt_stderr.log contains messages like this

2015-11-18 17:45:46.311 ShipIt[14118:552694] Beginning installation
2015-11-18 17:45:46.323 ShipIt[14118:552693] Couldn't abort install and restore owned bundle to previous location file:///Applications/Atom.app, error NSError { domain: NSPOSIXErrorDomain, code: 2, description: "No such file or directory" }
2015-11-18 17:45:48.215 ShipIt[14118:552695] Installation error: Error Domain=NSPOSIXErrorDomain Code=13 "Permission denied" UserInfo={NSLocalizedDescription=Permission denied}
2015-11-18 17:45:48.215 ShipIt[14118:552695] ShipIt quitting

joshaber:
⚠️ _Dependent on https://github.com/Squirrel/Squirrel.Mac/pull/206_ ⚠️ 

Fixes #157 
Fixes #131 

There were two problems:

1. We checked if we could write to the app bundle itself and escalated, but we didn't check if we could write to its parent directory to replace the bundle.
1. If we escalated, our user-scoped paths wouldn't be valid since ShipIt would run as root.

I still need to find a way to test this, even if just by hand, to verify the above issues really are fixed.


joshaber:
I've tested this manually with the following scenarios:

1. User-owned app bundle and parent directory. Succeeds 👍 
1. App bundle is owned by root. Squirrel escalates and succeeds 👍 
1. App bundle's parent directory is owned by root. Squirrel escalates and succeeds 👍

hhff:
Hi @joshaber - a big big thankyou for tackling this!  Really great stuff 👍

devtobo:
Is it fixed in Electron (not sure if they or I need to do something to have this fix for my users)?

joshaber:
No, Electron hasn't yet released an update which includes this fix.

Roverclover:
Y

Issue #64, opened 2013-10-04
Squirrel should support static update resources


keithduncan:
If we expect the server to branch on which response (update/no update) to send based on the version provided by the requester, whether that version is in a header or a query parameter, we can’t support static files.

So, is static file support important? If so we need to move the version equality check into the client so that it can retrieve the plain file and not update if the versions are the same. If it isn’t we need to remove that note from the docs and ensure that if the releases API supports Squirrel, it includes support for the same version comparison for Squirrel that Central does and Speakeasy will.

/cc @github/mac

gerges:
@dannygreg I know it's been some time, but what sort of long-term plans were you thinking? I'm thinking about tackling this now, since it seems like a tempting way to have a easy high-scale update distribution mechanism (CloudFront + S3)

bhajneet:
Can't we just format the jquery to say "No update" instead of relying on status codes?

Update:

```
{
    "update": "true",
    "url": "update.com/server/stuff"
}
```

No update:

```
{
    "update": "false"
}
```

romaincointepas:
It's also worth noting that Cloudflare does not allow caching of 204 responses (even with all the rules set to caching), meaning this issue also concerns standard non-static distribution model with Cloudflare on top.

PR #211, opened 2017-08-01
adding cdn-based updates: a file (simple text or json) containing rel…

tm243:
…ease information together with release files on a cdn


joshaber:
Thanks @tm243!

Supporting static update resources is something we've wanted for a while (see https://github.com/Squirrel/Squirrel.Mac/issues/64), so this is appealing.

That said, I think I need some more context around this to understand what problems this is solving and how it's solving them. There some changes here (a text format? another app?) whose purposes I don't fully understand.

felixrieseberg:
Thank you so much, this will be a gods-end for so many developers.

tm243:
We needed a solution at our company that allows us to roll out updates of electron apps over a standard http service rather than using the release server. It is easier to maintain and scales better. You can use any CDN and place your releases next to a release file containing some meta information.

The updater can read the RELEASES file which contains information about the latest release.
The file can be in json or plain text format, where the json format is similar to what the release server delivers. It comes with tests and a cli app to maintain the releases file. You can use it like this:

./SquirrelApp -releasify file.zip -version 0.0.143 -remote-path http://localhost/

and in your app:
http://localhost/releases/alpha/RELEASES.json?version=1.0.1&method=JSON

method can be JSON, Text, ReleaseServer.

The JSON format is basically a list of objects similar to what the release server delivers, and a pointer to the latest release (currentRelease), the text based file a list of releases with a timestamp. The ReleaseServer option is default, this way it shouldn't break anything (pls let me know if it does).

Happy to answer questions and/or modify the PR

joshaber:
Thanks! With that in mind, I have some more specific questions:

>The file can be in json or plain text format

Introducing a plain text format seems orthogonal to the goal and introduces new complexity. Why not continue to use our existing JSON format?

>a cli app to maintain the releases file

I'm still not sure I understand why there needs to be a CLI app? Could it be a separate project?

>The JSON format is basically a list of objects similar to what the release server delivers, and a pointer to the latest release

I'm not sure what roads you've already gone down, so apologies if this wouldn't work, but... what if, instead of treating 204 responses as "no update available" and 200 responses as "update available", we compared the `version` in the JSON response with the version of the app?

It seems like that'd give you the ability to use a static release file with minimal changes to Squirrel. Or I may be missing something 😄

develar:
Squirrel.Mac doesn't require special server. You can see solution in the https://www.npmjs.com/package/electron-simple-updater

Also, you can use https://www.npmjs.com/package/electron-updater 

electron-updater 1.x uses the same solution as electron-simple-updater (set feed url to special constructed json),
but electron-updater 2.x uses another solution to implement serverless updater — short-lived one-time proxy server. It allows us to implement progress events and unify publish providers (electron-updater supports s3, generic, github and so on).

I decided to not implement this logic in the Squirrel.Mac because:

* ability for end-users to use Electron Net API (e.g. intercept trafic, handle invalid/unsigned certs, proxy and so on).
* share code across platforms. No need to duplicate code for each platform. (That's why we have Electron).

So, it will be probably (electron-updater is tested and we don't have user reports that something wrong with implemented solution) cool to make some changes in the Squirrel.Mac to avoid need in the proxy server, but I doubt that we need to implement all features of electron-updater on Squirrel.Mac side. Better to implement only macOS specific native things.

develar:
For example, electron-updater approach allows us to provide united list of features — if something supported and tested for Windows (staged rollouts, custom provider), it will work for macOS as well (because native platform-specific code is responsible only for installing/applying downloaded update file).

MarshallOfSound:
@develar Let's keep on target here, don't forget that Electron is not the only consumer of `Squirrel.Mac/Windows` so this isn't just for Electron developers.  Also, the proxy server trick you reference is just a hack/workaround for the fact that static updates aren't supported yet.  This change will greatly improve the experience of any developer using Squirrel updaters.

develar:
@MarshallOfSound 

> Let's keep on target here
> roll out updates of electron app

Original author need this for Electron.

> the proxy server trick

No, it is required not for serverless. Squirrel.Mac supports static updates — just set feed url to special JSON file (this solution is used in the electron-updater 1.x and electron-simple-updater). Yes, not elegant, but works. 

I just to want say, that we should avoid some extra improvements in this project, because main consumers is Electron app (where we need to use cross-platform solution to avoid code duplication).
No doubt — simple way (without meta file) to use static files (http or file) is wanted.

tm243:
I'll take the text based non-json format out and make the cli app an external node app.

tm243:
I removed the SquirrelApp to maintain releases files and now its its own project here: https://github.com/tm243/Squirrel-cdn-cli  node/JS based

gerges:
This work was kicked-off with my recommendation so I feel somewhat responsible :-P. Zero intention of slowing this down, but thought it was worth mentioning: 

The introduction of the cli and the custom format was done with the intention of making this project more like Squirrel.Windows. I think we shouldn't underestimate the value of a more unified api between the platforms, even if we don't agree about the best technical implementation. Totally understand these projects aren't the same, but since they're simply called "Squirrel.Mac" and "Squirrel.Window" we were surprised to see how differently they work. I'd be incredibly happy to see this merged even without those additions, but I can't understate how the spirit of these changes mean less ad-hoc processes in our builds (updating release descriptors), the removal of mac only update servers, and all-in-all much less worrying about updates for us.

Squirrel/SQRLUpdater.m
```diff
@@ -150,18 +156,41 @@ - (id)initWithUpdateRequest:(NSURLRequest *)updateRequest {
...
+
+	//! download simple file
+
 	NSParameterAssert(updateRequest != nil);
 	NSParameterAssert(requestForDownload != nil);
 
 	self = [super init];
 	if (self == nil) return nil;
 
+	//! if ends with json, use the new way
```
joshaber: I'm not totally sure what this comment is referring to, but relying on the `json` extension wouldn't be quite right.

Squirrel/SQRLUpdater.m
```diff
@@ -150,18 +156,41 @@ - (id)initWithUpdateRequest:(NSURLRequest *)updateRequest {
 }
 
 - (id)initWithUpdateRequest:(NSURLRequest *)updateRequest requestForDownload:(SQRLRequestForDownload)requestForDownload {
+	//! foo.bar/releases.json?version=1.0
+	NSString *version = nil;
+	NSURLComponents* url = [[NSURLComponents alloc] initWithURL:updateRequest.URL resolvingAgainstBaseURL:FALSE];
+	NSArray* queryItems = url.queryItems;
+	Method mtd = RELEASESERVER;
```
joshaber: Could we name this variable `method` so it's clearer?

Squirrel/SQRLUpdater.m
```diff
@@ -150,18 +156,41 @@ - (id)initWithUpdateRequest:(NSURLRequest *)updateRequest {
...
+	for(NSURLQueryItem* obj in queryItems) {
```
joshaber: I'm not wild about switching update methods based on a query parameter. Is there a reason to do that over adding some new API to Squirrel.Mac to explicitly choose?

Squirrel/SQRLUpdater.m
```diff
@@ -210,14 +245,45 @@ - (id)initWithUpdateRequest:(NSURLRequest *)updateRequest requestForDownload:(SQ
...
+								// currentRelease is lower than version.
+								// Might be a new version for testing that is not deployed yet
+								// no roll back
+								NSLog(@"The running client is newer than the latest deployed release. Not downgrading.");
+								return [RACSignal empty];
+							}
+
+							//! @todo find latest
+							NSArray *releases = dict[@"releases"];
+							for(id obj in releases) {
```
joshaber: Can we give `obj` a type, even if it's just `NSDictionary`?

Squirrel/SQRLUpdater.m
```diff
@@ -493,8 +560,7 @@ - (RACSignal *)updateBundleMatchingCurrentApplicationInDirectory:(NSURL *)direct
 					NSLog(@"Could not open application bundle at %@", URL);
 					return NO;
 				}
-
-				return [bundle.bundleIdentifier isEqual:NSRunningApplication.currentApplication.bundleIdentifier];
+				return [bundle.bundleIdentifier isEqual:bundle.bundleIdentifier];
```
joshaber: This seems wrong / trivially true?

Squirrel/SQRLUpdater.m
```diff
@@ -382,6 +448,9 @@ - (RACSignal *)unarchiveAndPrepareData:(NSData *)data withName:(NSString *)name
 					NSError *error = nil;
 					if (![NSFileManager.defaultManager removeItemAtURL:zipOutputURL error:&error]) {
 						NSLog(@"Error removing downloaded archive at %@: %@", zipOutputURL, error.sqrl_verboseDescription);
+					} else {
+						[self updateBundleMatchingCurrentApplicationInDirectory:downloadDirectory];
```
joshaber: I'm not sure I understand this addition?

Squirrel/SQRLUpdater.m
tm243: yeah

Squirrel/SQRLUpdater.h
```diff
@@ -133,7 +152,7 @@ extern NSString * const SQRLUpdaterJSONObjectErrorKey;
 //                      updateRequest param.
 //
 // Returns the initialized `SQRLUpdater`.
-- (id)initWithUpdateRequest:(NSURLRequest *)updateRequest requestForDownload:(SQRLRequestForDownload)requestForDownload;
```
MarshallOfSound: We should leave this method format around as well to avoid this being a breaking change 👍

Squirrel/SQRLUpdater.m
```diff
@@ -149,17 +149,36 @@ - (id)initWithUpdateRequest:(NSURLRequest *)updateRequest {
...
+- (id)initWithUpdateRequest:(NSURLRequest *)updateRequest forVersion: (NSString*) version {
+	return [self initWithUpdateRequest:updateRequest requestForDownload:^(NSURL *downloadURL) {
+		return [NSURLRequest requestWithURL:downloadURL];
+	} forVersion:version useMode:JSONFILE];
+}
+
 - (id)initWithUpdateRequest:(NSURLRequest *)updateRequest requestForDownload:(SQRLRequestForDownload)requestForDownload {
+	return [self initWithUpdateRequest:updateRequest requestForDownload:^(NSURL *downloadURL) {
+		return [NSURLRequest requestWithURL:downloadURL];
+	} forVersion:@"" useMode:RELEASESERVER];
```
joshaber: Could we pass `nil` for version here, so that it's clear it's not meant to be used?

Squirrel/SQRLUpdater.m
```diff
@@ -149,17 +149,36 @@ - (id)initWithUpdateRequest:(NSURLRequest *)updateRequest {
...
+	if(mode == JSONFILE) {
+		NSParameterAssert(version != nil);
+	}
+
 	self = [super init];
 	if (self == nil) return nil;
 
 	_requestForDownload = [requestForDownload copy];
-	_updateRequest = [updateRequest copy];
+	_updateRequest = [NSURLRequest requestWithURL:updateRequest.URL cachePolicy:NSURLRequestReloadIgnoringCacheData timeoutInterval:60.0];
```
joshaber: What's the motivation for this change?

Squirrel/SQRLUpdater.m
```diff
@@ -216,8 +240,39 @@ - (id)initWithUpdateRequest:(NSURLRequest *)updateRequest requestForDownload:(SQ
...
+								NSLog(@"The running client is newer than the latest deployed release. Not downgrading.");
+								return [RACSignal empty];
+							}
+
+							//! @todo find latest
+							NSArray *releases = dict[@"releases"];
+							for(NSDictionary* release in releases) {
+								if([currentRelease isEqualToString:release[@"version"]]) {
+									bodyData = [[NSJSONSerialization dataWithJSONObject:release[@"updateTo"]
+																				options:0 error:&error] copy];
```
joshaber: Why this copy?

SquirrelTests/SQRLUpdaterSpec.m
```diff
@@ -12,12 +12,31 @@
...
 #import "SQRLZipArchiver.h"
 
 #import "OHHTTPStubs/OHHTTPStubs.h"
 #import "QuickSpec+SQRLFixtures.h"
 #import "SQRLTestUpdate.h"
 #import "TestAppConstants.h"
+#import <objc/objc-class.h>
+
+//! force Updater to believe we are not on readonly, to continue downloading the release
+bool isRunningOnReadOnlyVolumeImp(id self, SEL _cmd)
```
joshaber: Why do we need to swizzle this?

zeke: What does swizzle mean?

joshaber: Changing the implementation of a method for at runtime.

Squirrel/SQRLUpdater.m
```diff
@@ -216,8 +240,39 @@ - (id)initWithUpdateRequest:(NSURLRequest *)updateRequest requestForDownload:(SQ
 						NSError *error = [NSError errorWithDomain:SQRLUpdaterErrorDomain code:SQRLUpdaterErrorReadOnlyVolume userInfo:errorInfo];
 						return [RACSignal error:error];
 					}
-				}
 
+					if(mode == JSONFILE) {
+						NSError *error = nil;
+						NSDictionary *dict = [NSJSONSerialization JSONObjectWithData:bodyData options:0 error:&error];
```
joshaber: It doesn't look like we're actually checking for failing to parse the JSON.

Squirrel/SQRLUpdater.m
```diff
@@ -216,8 +240,39 @@ - (id)initWithUpdateRequest:(NSURLRequest *)updateRequest requestForDownload:(SQ
...
+								// no roll back
+								NSLog(@"The running client is newer than the latest deployed release. Not downgrading.");
+								return [RACSignal empty];
+							}
+
+							//! @todo find latest
+							NSArray *releases = dict[@"releases"];
+							for(NSDictionary* release in releases) {
+								if([currentRelease isEqualToString:release[@"version"]]) {
+									bodyData = [[NSJSONSerialization dataWithJSONObject:release[@"updateTo"]
```
joshaber: Same as above, it doesn't look like we're actually checking for failing to parse the JSON.

zeke:
I'm really glad to see this inching forward. Will be a big win if we can ship it.

joshaber:
Superseded by #219.

PR #219, opened 2018-01-04
Add CDN based releases

### LINKED ISSUE #64: Squirrel should support static update resources

MarshallOfSound:
Fixes #64 

This PR supersedes #211 and contains just a few extra fixes on top of it to satisfy the outstanding comments from @joshaber


MarshallOfSound:
Please note the only place this touches the existing code path is the modification of the cache policy  and timeout.  Neither of which should have any negative impacts on existing users.

Squirrel/SQRLUpdater.m
```diff
@@ -149,17 +149,43 @@ - (id)initWithUpdateRequest:(NSURLRequest *)updateRequest {
...
 	if (self == nil) return nil;
 
 	_requestForDownload = [requestForDownload copy];
-	_updateRequest = [updateRequest copy];
-	_updateClass = SQRLUpdate.class;
+	NSMutableURLRequest* mutableUpdateRequest = [updateRequest mutableCopy];
 
+	if (mode == JSONFILE) {
+		mutableUpdateRequest.cachePolicy = NSURLRequestReloadIgnoringCacheData;
+		mutableUpdateRequest.timeoutInterval = 60.0;
```
joshaber: Is there a reason to set this so high? Can we rely on users to set this high if needed?

Squirrel/SQRLUpdater.m
```diff
@@ -149,17 +149,43 @@ - (id)initWithUpdateRequest:(NSURLRequest *)updateRequest {
...
+	}
+	_updateRequest = mutableUpdateRequest;
+	_updateRequest = [NSURLRequest requestWithURL:updateRequest.URL cachePolicy:NSURLRequestReloadIgnoringCacheData timeoutInterval:60.0];
```
joshaber: I'm not sure I understand why we're assigning `_updateRequest` to `mutableUpdateRequest` and then doing this too?

Squirrel/SQRLUpdater.m
```diff
@@ -216,8 +247,58 @@ - (id)initWithUpdateRequest:(NSURLRequest *)updateRequest requestForDownload:(SQ
...
+
+						NSString *currentRelease = dict[@"currentRelease"];
+						if(currentRelease) {
+							//! if CDN points to the currently running version as the latest version, bail out
+							if([currentRelease isEqualToString:version]) {
+								NSLog(@"The running client is already the latest version.");
+								return [RACSignal empty];
+							}
+
+							if ([version compare:currentRelease options:NSNumericSearch] == NSOrderedDescending) {
```
joshaber: Does this order properly with versions like `1.2.3` vs. `1.2.4`, or do version have to be integers?

Squirrel/SQRLUpdater.h
```diff
@@ -67,6 +67,12 @@ extern NSString * const SQRLUpdaterJSONObjectErrorKey;
 @class RACDisposable;
 @class RACSignal;
 
+/// Type of mode used to download the release
+typedef enum {
+	JSONFILE=1,
```
joshaber: Could we document the expected format somewhere? In the readme, probably?

Squirrel/SQRLUpdater.m
```diff
@@ -216,8 +247,58 @@ - (id)initWithUpdateRequest:(NSURLRequest *)updateRequest requestForDownload:(SQ
 						NSError *error = [NSError errorWithDomain:SQRLUpdaterErrorDomain code:SQRLUpdaterErrorReadOnlyVolume userInfo:errorInfo];
 						return [RACSignal error:error];
 					}
-				}
 
+					if(mode == JSONFILE) {
```
joshaber: It seems like this should be an `else if` on the `if(mode == RELEASESERVER)` branch above?

SquirrelTests/SQRLUpdaterSpec.m
```diff
@@ -59,6 +78,134 @@
...
+			//! setup the updater
+			localRequest = [NSURLRequest requestWithURL:[NSURL URLWithString:localServerURL]];
+			updater = [[SQRLUpdater alloc] initWithUpdateRequest:localRequest];
+			
+			//! replace updateFromJSONData() and isRunningOnReadOnlyVolume() methods
+			method_setImplementation(class_getInstanceMethod([SQRLUpdater class]
+															 , @selector(updateFromJSONData:))
+									 , (IMP) updateFromJSONDataImp);
+			
+			method_setImplementation(class_getInstanceMethod([SQRLUpdater class]
```
joshaber: Why do we need to swizzle these methods?

Squirrel/SQRLUpdater.m
MarshallOfSound: Yes, I wasn't convinced at first either but it correctly sorts versions with major.minor.patch

Squirrel/SQRLUpdater.m
MarshallOfSound: Heh, woops

MarshallOfSound:
Submodule changes are being reverted 👍

SquirrelTests/SQRLUpdaterSpec.m
tm243: the methods are mocked basically. We don't want updateFromJSONData to be executed. It is enough to know if it was called

Squirrel/SQRLUpdater.m
MarshallOfSound: This is actually the default for the default cache policy.  The apple docs don't specify the default for this cache policy so we set it explicitly to be safe 👍

MarshallOfSound:
@joshaber Feedback addressed, the JSON structure is now in the README, it is also in the test docs 👍

Squirrel/SQRLUpdater.m
joshaber: Ah I see. Doesn't this mean that users can't customize it, since we're always set it to 60?

Squirrel/SQRLUpdater.m
```diff
@@ -216,8 +230,72 @@ - (id)initWithUpdateRequest:(NSURLRequest *)updateRequest requestForDownload:(SQ
...
+								if([currentRelease isEqualToString:release[@"version"]]) {
+									bodyData = [NSJSONSerialization dataWithJSONObject:release[@"updateTo"]
+																				options:0 error:&error];
+									break;
+								}
+							}
+						}
+					}
+				}
+				if (bodyData == nil) {
```
joshaber: I'm not totally sure I understand the need for this branch. By this point we've already tried to use `bodyData`, so won't we already have run into errors?

Squirrel/SQRLUpdater.m
joshaber: Nice!

Squirrel/SQRLUpdater.m
MarshallOfSound: This handles the case where the JSON is malformed and `updateTo` isn't set

Squirrel/SQRLUpdater.m
MarshallOfSound: Yeah, happy to remove this if we want to allow users to customize

tm243: In that case, to be consistent, it should be configurable for server-mode as well. As seen here: https://github.com/tm243/Squirrel.Mac/commit/602bb558d56f3867ab8077f24fc4dd40c4c08777
The apple doc says it defaults to 60 seconds, but in older iOS versions the default value was 240 seconds.

Squirrel/SQRLUpdater.m
tm243: https://github.com/tm243/Squirrel.Mac/commit/38411882b46d5be7438f2ee89c86555886222340

Squirrel/SQRLUpdater.m
joshaber: It's already configurable, isn't it? By virtue of the user passing in `updateRequest`?

Squirrel/SQRLUpdater.m
joshaber: Hmm, I see. In that case I think we should do this check closer to line 282, since it seems specific to that.

MarshallOfSound: There are other escape conditions earlier than that for loop that we need to check (by checking for `nil`).  This is the best place for it to go so that we don't duplicate the error checking code any more than we need to.

MarshallOfSound:
Timeout removed and nil check comment addressed

Squirrel/SQRLUpdater.m
tm243: I agree here.
This hasn't been checked so far, so in the context of this PR it would be Ok to do this:
https://github.com/Squirrel/Squirrel.Mac/pull/211/commits/9efd0ed4a85c2c01c7c655fbd8f78b3fb4a95199
but I think its a good idea to have that check for both modes.

zeke:
Awesome!

What's the next step? Do we need to update Squirrel in https://github.com/electron/electron-frameworks ?

joshaber:
Yup! I'll make a release of Squirrel next week and then we'll need to make a new release with that.

paulcbetts:
This is the Hero We Deserve™

zeke:
@joshaber is the Squirrel release process documented somewhere? Maybe I can take a shot at it?

joshaber:
Sorry for the delay! It's not documented, but _in theory_ it's `carthage build --platform macos --no-skip-current && carthage archive`. Lemme know if you run into any trouble @zeke. I'm keen to enable you all as much as possible 😄

zeke:
Ok, first question: what's `carthage` :)

joshaber:
It's a dependency manager for Cocoa: https://github.com/Carthage/Carthage. You _should_ be able to `brew install carthage`.

MarshallOfSound:
It's a bit of a pain to build Squirrel.Mac locally, you need to be running an old XCode version 😄   Here is a build from the latest master 👍 

[Squirrel.framework.zip](https://github.com/Squirrel/Squirrel.Mac/files/1697854/Squirrel.framework.zip)

PR #231, opened 2018-10-26
fix jump link to update-file-json-format

olegsklyarov:
Fix broken link to update-file-json-format header inside README


joshaber:
Thank you!

PR #236, opened 2019-04-03
Fix ShipIt dependency linking.

devinfoley:
A recent change to Squirrel attempted to link to ReactiveCocoa and Mantle in ShipIt, vs copying and compiling their source files into the ShipIt binary. This was done because, in certain environments, dylib was loading multiple copies of the classes, which was causing errors in ReactiveCocoa's assertions.

This PR actually removes the dependency classes from the Copy Files build step. It adds an additional build step to re-link ReactiveCocoa and Mantle, using @executable_path instead of @rpath, as @rpath linking was causing the same double loading problem in certain environments.

### DIFF
--- a/Squirrel.xcodeproj/project.pbxproj
+++ b/Squirrel.xcodeproj/project.pbxproj
@@ -7,8 +7,8 @@
 	objects = {
 
 /* Begin PBXBuildFile section */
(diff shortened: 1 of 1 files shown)


Neoster:
When will this be merged? I think it will fix my issues i am having using Squirrel....

devinfoley:
Closing because this is covered by #242

PR #249, opened 2020-06-29
chore: fix missing deallocs, deprecation warnings and invalid casts

MarshallOfSound:
Found while building with Chromiums config.  Couple of missing `super dealloc` calls, an invalid literal cast and some deprecation warnings which we can't do anything about so let's just suppress them


devinfoley:
:shipit:

PR #253, opened 2020-09-28
fix: use new ShipIt binary on macOS 11.0 to workaround flaw in launchApplicationAtURL

MarshallOfSound:
It looks like `launchApplicationAtURL` doesn't work in background tasks on Big Sur (nor do any of the 4 alternative APIs).  This works around an issue where the updated app wouldn't launch on 11.0 after the update.  This issue has been reported to apple and we hope to remove this workaround soon.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,4 @@
-osx_image: xcode8.3
+osx_image: xcode9.4
 language: objective-c
 
 install: script/bootstrap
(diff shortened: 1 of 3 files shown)


kklash (approved):

MarshallOfSound:
Closing out as this is a super-flaky solution that we kinda don't want

MarshallOfSound:
Re-opening as we have a modified version of this that appears to work reliably

devinfoley (approved):

PR #255, opened 2020-10-15
fix: Ensure we use a writable log file.

devinfoley:
Squirrel.mac has an issue wherein if ShipIt_stderr.log or ShipIt_stdout.log aren't writable, ShipIt will fail to launch. If the entire ShipIt directory isn't writable, we detect that and prompt the user for credentials, but if the directory is writable and the log files contained are not, we just die silently.

This PR changes the logic used the generate the log file URLs so that we always generate writable log file URLs. Before returning the log file URL, we test to see if it's writable. If not, we append a number to the URL and test that URL. We repeat this process until we find a writable log file URL, or we reach 100 and give up.

I also included a little code cleanup to reduce duplication, so that this new logic doesn't get copied all over the place.

### DIFF
--- a/Squirrel/SQRLDirectoryManager.m
+++ b/Squirrel/SQRLDirectoryManager.m
@@ -66,31 +66,52 @@
 		setNameWithFormat:@"%@ -storageURL", self];
 }
 
-- (RACSignal *)shipItStateURL {
+// Returns a signal to a URL with the given filename and job name inside of the storage folder
+//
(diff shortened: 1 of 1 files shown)


MarshallOfSound (commented):
Looks good, maybe add some logging around failure to write files?  E.g. If we get past 100 we should put something in the logs

devinfoley:
@MarshallOfSound good call. Added!

MarshallOfSound (approved):

PR #295, opened 2026-05-02
Update test dependencies for Xcode 15+ compatibility

augiekim:
Quick 1.x and Nimble 6.x require Swift 3.0, which was dropped in Xcode 15. Update to Quick 5.0.1 + Nimble 9.2.1 — the last versions that maintain full backward compatibility with the Objective-C QuickSpec DSL used throughout the test suite (QuickSpecBegin/End, it, describe, beforeEach, etc.).

- Cartfile.private: bump Quick ~> 5.0, Nimble ~> 9.0
- Cartfile.resolved: pin Quick v5.0.1, Nimble v9.2.1
- Squirrel.xcodeproj: set SWIFT_VERSION = 5.0 (was 3.0) so SwiftSpec.swift and the test target build cleanly under the modern Swift compiler

### DIFF
--- a/Cartfile.private
+++ b/Cartfile.private
@@ -1,3 +1,3 @@
 github "jspahrsummers/xcconfigs" ~> 0.8.1
-github "Quick/Quick" ~> 1.1.0
-github "Quick/Nimble" ~> 6.1.0
+github "Quick/Quick" ~> 5.0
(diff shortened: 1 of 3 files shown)


MarshallOfSound:
I think superseded by #296

PR #324, opened 2026-07-21
fix: bound zip archiver standard error collection

KevinWang-wpq:
## Summary

- Replace unbounded `RACSignal collect` stderr accumulation with a streaming
  aggregate that retains at most 1 MiB per `ditto` invocation.
- Stop the readability handler after EOF, avoiding unnecessary empty-pipe work.
- Add a regression test that makes a failing task emit 2 MiB on stderr and
  verifies that the retained error text is capped.

## Root cause

`SQRLZipArchiver` forwards every `NSFileHandle.availableData` callback into a
`RACSubject`. `collect` retains every chunk in an `NSMutableArray`, after which
the code concatenates all chunks into another `NSMutableData`. A malformed
archive can make `ditto` write excessive diagnostics and cause an unbounded
allocation in the host process.

### Observed crash path (root-cause-relevant frames)

```
[Chromium / PartitionAlloc
 base/allocator/partition_allocator]
partition_alloc::internal::OnNoMemoryInternal
partition_alloc::internal::TerminateBecauseOutOfMemory
partition_alloc::internal::OnNoMemory
partition_alloc::internal::PartitionExcessiveAllocationSize
partition_alloc::internal::PartitionDirectMap

[Chromium / allocator shim]
allocator shim allocation path

[Apple Foundation]
-[__NSArrayM insertObject:atIndex:]

[ReactiveObjC
 ReactiveObjC/RACSignal+Operations.m]
-[RACSignal collect]

[Squirrel.Mac
 Squirrel/SQRLZipArchiver.m]
__23-[SQRLZipArchiver init]_block_invoke.44

[Apple Foundation]
-[NSConcreteFileHandle _monitor]
```

`PartitionExcessiveAllocationSize` is a deliberate PartitionAlloc termination
when one allocation exceeds the direct-map limit (roughly 2 GiB), rather than a
generic low-memory termination. Here the unbounded `collect` array is the
allocation owner; the Foundation frame is where that array growth reaches the
allocator.

The error output is used only to populate `NSLocalizedDescription` when the
`ditto` task exits unsuccessfully. Exit status and normal archive behavior are
unchanged. The first 1 MiB of diagnostics is retained; later data is still read
and discarded so the child process cannot block on a full stderr pipe.

### DIFF
--- a/Squirrel/SQRLZipArchiver.m
+++ b/Squirrel/SQRLZipArchiver.m
@@ -14,6 +14,11 @@ NSString * const SQRLZipArchiverErrorDomain = @"SQRLZipArchiverErrorDomain";
 NSString * const SQRLZipArchiverExitCodeErrorKey = @"SQRLZipArchiverExitCodeErrorKey";
 const NSInteger SQRLZipArchiverShellTaskFailed = 1;
 
+// `ditto` writes diagnostics for input it cannot process. Keep the retained
+// untrusted output bounded so a failed archive operation cannot exhaust memory.
+static const NSUInteger kSQRLZipArchiverMaximumStandardErrorDataLength =
+	1024 * 1024;
+
 @interface SQRLZipArchiver () {
 	RACSubject *_taskTerminated;
 }
@@ -30,8 +35,8 @@ const NSInteger SQRLZipArchiverShellTaskFailed = 1;
 // A pipe used for reading error logging from `dittoTask`.
 @property (nonatomic, strong, readonly) NSPipe *standardErrorPipe;
 
-// Sends an NSData representing the error logging from `dittoTask` once the task
-// has terminated.
+// Sends up to 1 MiB of error logging from `dittoTask` once the task has
+// terminated.
 @property (nonatomic, strong, readonly) RACSignal *standardErrorData;
 
 // Launches the receiver's `dittoTask` with the given command line arguments.
@@ -68,19 +73,29 @@ const NSInteger SQRLZipArchiverShellTaskFailed = 1;
 
 	RACSubject *errorDataChunks = [[RACSubject subject] setNameWithFormat:@"errorDataChunks"];
 	self.standardErrorPipe.fileHandleForReading.readabilityHandler = ^(NSFileHandle *handle) {
-		[errorDataChunks sendNext:handle.availableData];
+		NSData *data = handle.availableData;
+		if (data.length == 0) {
+			handle.readabilityHandler = nil;
+			return;
+		}
+
+		[errorDataChunks sendNext:data];
 	};
 
-	_standardErrorData = [[[[[[errorDataChunks
+	_standardErrorData = [[[[[errorDataChunks
 		takeUntil:self.taskTerminated]
-		collect]
-		flattenMap:^(NSArray *chunks) {
-			NSMutableData *combined = [NSMutableData data];
-			for (NSData *data in chunks) {
-				[combined appendData:data];
+		aggregateWithStartFactory:^id {
+			return [NSMutableData data];
+		} reduce:^id(NSMutableData *combined, NSData *data) {
+			if (combined.length >= kSQRLZipArchiverMaximumStandardErrorDataLength || data.length == 0) {
+				return combined;
 			}
 
-			return [RACSignal return:combined];
+			NSUInteger remainingLength =
+				kSQRLZipArchiverMaximumStandardErrorDataLength - combined.length;
+			NSUInteger appendLength = MIN(data.length, remainingLength);
+			[combined appendBytes:data.bytes length:appendLength];
+			return combined;
 		}]
 		repeat]
 		takeUntil:self.rac_willDeallocSignal]
--- a/Squirrel/SQRLZipArchiver.m
+++ b/Squirrel/SQRLZipArchiver.m
@@ -16,8 +16,29 @@ const NSInteger SQRLZipArchiverShellTaskFailed = 1;
 
 // `ditto` writes diagnostics for input it cannot process. Keep the retained
 // untrusted output bounded so a failed archive operation cannot exhaust memory.
-static const NSUInteger kSQRLZipArchiverMaximumStandardErrorDataLength =
-	1024 * 1024;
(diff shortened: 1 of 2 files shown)


KevinWang-wpq:
Hi @MarshallOfSound — this bounds unbounded stderr collection in SQRLZipArchiver and adds a regression test for excessive diagnostic output. Since you recently updated this file in #308, could you please take a look when you have a chance?

Squirrel/SQRLZipArchiver.m
```diff
@@ -14,6 +14,11 @@
 NSString * const SQRLZipArchiverExitCodeErrorKey = @"SQRLZipArchiverExitCodeErrorKey";
 const NSInteger SQRLZipArchiverShellTaskFailed = 1;
 
+// `ditto` writes diagnostics for input it cannot process. Keep the retained
+// untrusted output bounded so a failed archive operation cannot exhaust memory.
+static const NSUInteger kSQRLZipArchiverMaximumStandardErrorDataLength =
+	1024 * 1024;
```
MarshallOfSound: No `k` prefix elsewhere in this repo (e.g. `SQRLShipItMaximumInstallationAttempts`), and this fits on one line
```suggestion
static const NSUInteger SQRLZipArchiverMaximumStandardErrorDataLength = 1024 * 1024;
```

Squirrel/SQRLZipArchiver.m
```diff
@@ -68,19 +73,29 @@ - (id)init {
...
 		takeUntil:self.taskTerminated]
-		collect]
-		flattenMap:^(NSArray *chunks) {
-			NSMutableData *combined = [NSMutableData data];
-			for (NSData *data in chunks) {
-				[combined appendData:data];
+		aggregateWithStartFactory:^id {
+			return [NSMutableData data];
+		} reduce:^id(NSMutableData *combined, NSData *data) {
+			if (combined.length >= kSQRLZipArchiverMaximumStandardErrorDataLength || data.length == 0) {
```
MarshallOfSound: Can't hit this anymore since the readability handler filters empty reads before `sendNext:` — can drop it
```suggestion
			if (combined.length >= kSQRLZipArchiverMaximumStandardErrorDataLength) {
```

Squirrel/SQRLZipArchiver.m
```diff
@@ -68,19 +73,29 @@ - (id)init {
...
+				return combined;
 			}
 
-			return [RACSignal return:combined];
+			NSUInteger remainingLength =
+				kSQRLZipArchiverMaximumStandardErrorDataLength - combined.length;
+			NSUInteger appendLength = MIN(data.length, remainingLength);
+			[combined appendBytes:data.bytes length:appendLength];
```
MarshallOfSound, line 122: Truncating at an arbitrary byte can split a multibyte UTF-8 sequence (ditto echoes file paths so non-ASCII is plausible), and then `initWithData:encoding:NSUTF8StringEncoding` down in `-launchWithArguments:` returns nil and we lose the whole description. We should handle it here since the cap is what introduces it — rather than a decode-retry loop, when we actually cut mid-chunk we can walk the cut point back to a code-point boundary, it's at most 3 bytes since UTF-8 self-syncs. Something like a small `SQRLUTF8TruncationBoundary(data.bytes, appendLength)` helper that skips back over continuation bytes and drops the incomplete sequence, called only when `appendLength < data.length`

SquirrelTests/SQRLZipArchiverSpec.m
```diff
@@ -64,6 +64,26 @@ - (RACSignal *)launchWithArguments:(NSArray *)arguments;
...
+			@"/usr/bin/yes x | /usr/bin/head -c 2097152 >&2; exit 1",
+		]]
+		asynchronouslyWaitUntilCompleted:&error];
+
+	expect(@(success)).to(beFalsy());
+	expect(error.domain).to(equal(SQRLZipArchiverErrorDomain));
+
+	NSString *errorString = error.userInfo[NSLocalizedDescriptionKey];
+	expect(errorString).notTo(beNil());
+	expect(@(errorString.length)).to(beLessThanOrEqualTo(@(1024 * 1024)));
```
MarshallOfSound, line 89: Should we also assert a lower bound? As-is this passes trivially if the description ends up empty/nil, so it doesn't prove the cap is what's binding
```suggestion
	expect(@(errorString.length)).to(beGreaterThan(@(512 * 1024)));
	expect(@(errorString.length)).to(beLessThanOrEqualTo(@(1024 * 1024)));
```

MarshallOfSound (commented):
Verified locally, `SQRLZipArchiverSpec` all green including the new test. Nice — factory variant of aggregate is the right call given the `repeat`, and killing the readabilityHandler on EOF fixes the post-EOF spin as a bonus. Few things inline.

KevinWang-wpq:
@MarshallOfSound Thanks again for the review — I have addressed all of the inline suggestions in commit 0a8ae53.

- aligned the constant naming with the repository convention;
- removed the unreachable empty-data guard;
- made truncation preserve a valid UTF-8 boundary and stop accepting subsequent chunks once capped;
- strengthened the regression test with a UTF-8 boundary case and a lower-bound assertion.

script/test -derivedDataPath DerivedData passes (88 tests). Could you please take another look when you have a chance?

KevinWang-wpq:
@MarshallOfSound I wanted to get your take on a simpler alternative for the UTF-8 handling before changing the current implementation.

Instead of finding a UTF-8 boundary while accumulating stderr, we could keep the original bounded NSMutableData accumulator unchanged. At the final decode point, attempt UTF-8 decoding normally; only if it returns nil, retry after removing 1, then 2, then 3 bytes from the end.

Because the cap only slices the tail of the byte stream, any new invalid UTF-8 introduced by the cap should be confined to that suffix; a UTF-8 scalar is at most 4 bytes. This keeps the 1 MiB cap behavior unchanged and avoids the byte-mask helper plus the extra truncated state in the accumulator. The retry would only run once on the task-failure path.

I realize your inline suggestion preferred boundary handling over a decode-retry loop, so I have not changed the implementation to this approach. Would you be open to this tradeoff, or would you prefer to keep the current boundary-at-truncation approach?

PR #327, opened 2026-08-11
feat: stream update downloads to disk, resume them, and verify a declared digest

codebytere:
The update ZIP is fetched with a single `NSURLConnection sendAsynchronousRequest:` and held as one `NSData` before it is written out, so a large update is an equally large allocation in the host app, and any interruption (network change, sleep, quit) discards the whole transfer until the next check starts it again from zero. `NSURLConnection` has also been deprecated since 10.11.

This adds `SQRLDownloader`, a small wrapper around an `NSURLSession` download task, and moves `-downloadBundleForUpdate:intoDirectory:` onto it:

- the payload streams to a file in the update directory; nothing is buffered in memory
- when a transfer fails or is cancelled and the response carried a validator, the session's resume data is kept beside the `update.XXXXXXX` directories (`download.resumedata`), and the next check continues from that offset with a `Range` request instead of starting over; this holds after a relaunch too, since downloads still in flight at `NSApplicationWillTerminateNotification` hand over their resume data first
- resume data for a different URL, or that the session no longer accepts, is dropped and the download starts fresh, so the worst case is today's behaviour
- progress is published as `SQRLDownloadProgress` values on the new `SQRLUpdater.downloadProgress`
- `SQRLDownloader.sessionConfiguration` lets a host set network policy or protocol classes for the download sessions

Separately, an update may now declare `sha256` and `size` for its ZIP (`SQRLUpdate.packageDigest` / `packageSize`, both optional). A download that does not match is deleted before `ditto` runs and the check errors with `SQRLUpdaterErrorInvalidUpdatePackage`. Signature verification and installation are untouched.

Tests: `SQRLDownloaderSpec` runs against a small local HTTP server (`SquirrelTests/TestServer.py`, ranges + ETag, can cut a transfer short) and covers a full download with progress, resuming an interrupted transfer (asserts the `Range` request and a byte-identical result), keeping resume data on disposal and on `+cancelAllWritingResumeDataWithTimeout:`, and discarding resume data for another URL. `SQRLUpdaterSpec` gains the digest/size rejection case and `SQRLUpdateSpec` the parsing; the existing out-of-process update specs exercise the new path end to end over `file://`. 94 tests pass locally on macOS 26.5 / Xcode 26.5

### DIFF
--- a/Squirrel/SQRLDownloader.m
+++ b/Squirrel/SQRLDownloader.m
@@ -49,10 +49,11 @@ static NSString * const SQRLDownloaderResumeDataKey = @"resumeData";
 @property (nonatomic, strong) id<RACSubscriber> subscriber;
 @property (nonatomic, copy) NSURL *destinationURL;
 @property (nonatomic, strong) NSError *moveError;
+// The resume data this downloader took off disk, in case it must go back.
+@property (nonatomic, copy) NSData *takenResumeData;
 
 @property (nonatomic, assign) int64_t bytesResumed;
 @property (nonatomic, assign) BOOL startedFromResumeData;
-@property (nonatomic, assign) BOOL retriedFresh;
 @property (nonatomic, assign) BOOL receivedBytes;
 @property (nonatomic, assign) BOOL cancelled;
 
@@ -212,6 +213,7 @@ static NSURLSessionConfiguration *SQRLDownloaderSessionConfiguration = nil;
 	dispatch_group_enter(self.completionGroup);
 
 	NSURLSessionDownloadTask *resumed = resumeData != nil ? [self.session downloadTaskWithResumeData:resumeData] : nil;
+	if (resumed != nil) self.takenResumeData = resumeData;
 	self.startedFromResumeData = (resumed != nil);
 	self.receivedBytes = NO;
 	self.bytesResumed = 0;
@@ -268,19 +270,17 @@ static NSURLSessionConfiguration *SQRLDownloaderSessionConfiguration = nil;
 	NSInteger statusCode = [task.response isKindOfClass:NSHTTPURLResponse.class] ? [(NSHTTPURLResponse *)task.response statusCode] : 200;
 	BOOL resumeWentNowhere = self.startedFromResumeData && !self.cancelled && (error != nil ? !self.receivedBytes : statusCode < 200 || statusCode > 299);
 	if (resumeWentNowhere) {
-		if (!self.retriedFresh) {
-			self.retriedFresh = YES;
-			dispatch_group_leave(self.completionGroup);
-			[self startTaskWithResumeData:nil];
-			return;
-		}
-		error = error ?: [NSError errorWithDomain:NSURLErrorDomain code:NSURLErrorCannotLoadFromNetwork userInfo:@{ NSURLErrorKey: self.request.URL }];
-		[self finishWithError:error];
+		dispatch_group_leave(self.completionGroup);
+		[self startTaskWithResumeData:nil];
 		return;
 	}
 
 	if (error != nil) {
-		[self storeResumeData:error.userInfo[NSURLSessionDownloadTaskResumeData]];
+		// A fresh attempt that received nothing either means the network is
+		// the problem, not the ranged request: put back what was taken so
+		// the partial file outlives the outage.
+		NSData *resumeData = error.userInfo[NSURLSessionDownloadTaskResumeData];
+		[self storeResumeData:(!self.receivedBytes && self.takenResumeData != nil) ? self.takenResumeData : resumeData];
 		[self finishWithError:error];
 		return;
 	}
--- a/Squirrel/SQRLDownloader.m
+++ b/Squirrel/SQRLDownloader.m
@@ -53,6 +53,7 @@ static NSString * const SQRLDownloaderResumeDataKey = @"resumeData";
 @property (nonatomic, assign) int64_t bytesResumed;
 @property (nonatomic, assign) BOOL startedFromResumeData;
 @property (nonatomic, assign) BOOL retriedFresh;
+@property (nonatomic, assign) BOOL receivedBytes;
 @property (nonatomic, assign) BOOL cancelled;
 
 // Entered when a task starts, left when its completion (including any resume
@@ -144,10 +145,12 @@ static NSURLSessionConfiguration *SQRLDownloaderSessionConfiguration = nil;
 
 	NSDictionary *stored = [NSDictionary dictionaryWithContentsOfURL:self.resumeDataURL];
 	NSData *resumeData = stored[SQRLDownloaderResumeDataKey];
-	if (![stored[SQRLDownloaderResumeURLKey] isEqual:self.request.URL.absoluteString] || ![resumeData isKindOfClass:NSData.class]) {
-		[self clearStoredResumeData];
-		return nil;
-	}
+
+	// Taken, not read: while this attempt owns the partial file nothing on disk
+	// may point another downloader (or a relaunch after a crash) at it. It is
+	// written again if this attempt stops short.
+	[self clearStoredResumeData];
+	if (![stored[SQRLDownloaderResumeURLKey] isEqual:self.request.URL.absoluteString] || ![resumeData isKindOfClass:NSData.class]) return nil;
 
 	return resumeData;
 }
@@ -208,10 +211,12 @@ static NSURLSessionConfiguration *SQRLDownloaderSessionConfiguration = nil;
 - (void)startTaskWithResumeData:(NSData *)resumeData {
 	dispatch_group_enter(self.completionGroup);
 
-	self.startedFromResumeData = (resumeData != nil);
+	NSURLSessionDownloadTask *resumed = resumeData != nil ? [self.session downloadTaskWithResumeData:resumeData] : nil;
+	self.startedFromResumeData = (resumed != nil);
+	self.receivedBytes = NO;
 	self.bytesResumed = 0;
 	self.moveError = nil;
-	self.task = self.startedFromResumeData ? [self.session downloadTaskWithResumeData:resumeData] : [self.session downloadTaskWithRequest:self.request];
+	self.task = resumed ?: [self.session downloadTaskWithRequest:self.request];
 	[self.task resume];
 }
 
@@ -238,6 +243,7 @@ static NSURLSessionConfiguration *SQRLDownloaderSessionConfiguration = nil;
 }
 
 - (void)URLSession:(NSURLSession *)session downloadTask:(NSURLSessionDownloadTask *)downloadTask didWriteData:(int64_t)bytesWritten totalBytesWritten:(int64_t)totalBytesWritten totalBytesExpectedToWrite:(int64_t)totalBytesExpectedToWrite {
+	self.receivedBytes = YES;
 	[self sendProgressWithBytesReceived:totalBytesWritten bytesExpected:totalBytesExpectedToWrite];
 }
 
@@ -254,19 +260,27 @@ static NSURLSessionConfiguration *SQRLDownloaderSessionConfiguration = nil;
 - (void)URLSession:(NSURLSession *)session task:(NSURLSessionTask *)task didCompleteWithError:(NSError *)error {
 	if (task != self.task) return;
 
-	if (error != nil) {
-		NSData *resumeData = error.userInfo[NSURLSessionDownloadTaskResumeData];
-		[self storeResumeData:resumeData];
-
-		// Resume data the session refused (stale temporary file, format it no
-		// longer reads): don't lose this attempt to it, start over once.
-		if (resumeData == nil && self.startedFromResumeData && !self.retriedFresh && !self.cancelled) {
+	// A resumed attempt that got no further than where it started says nothing
+	// about the payload: the session refused the data, the ranged request was
+	// reset or could not be decoded, or it was answered rather than served
+	// (403/416 from an expired signed URL, 304). Forget the resume data and
+	// ask from the start, once; never hand such an answer up as the download.
+	NSInteger statusCode = [task.response isKindOfClass:NSHTTPURLResponse.class] ? [(NSHTTPURLResponse *)task.response statusCode] : 200;
+	BOOL resumeWentNowhere = self.startedFromResumeData && !self.cancelled && (error != nil ? !self.receivedBytes : statusCode < 200 || statusCode > 299);
+	if (resumeWentNowhere) {
+		if (!self.retriedFresh) {
 			self.retriedFresh = YES;
 			dispatch_group_leave(self.completionGroup);
 			[self startTaskWithResumeData:nil];
 			return;
 		}
+		error = error ?: [NSError errorWithDomain:NSURLErrorDomain code:NSURLErrorCannotLoadFromNetwork userInfo:@{ NSURLErrorKey: self.request.URL }];
+		[self finishWithError:error];
+		return;
+	}
 
+	if (error != nil) {
+		[self storeResumeData:error.userInfo[NSURLSessionDownloadTaskResumeData]];
 		[self finishWithError:error];
 		return;
 	}
--- a/Squirrel/SQRLDownloader.m
+++ b/Squirrel/SQRLDownloader.m
@@ -1,0 +1,305 @@
+//
+//  SQRLDownloader.m
+//  Squirrel
+//
+//  Copyright (c) 2026 GitHub. All rights reserved.
+//
+
+#import "SQRLDownloader.h"
+#import <ReactiveObjC/ReactiveObjC.h>
+
+// Keys of the property list written to `resumeDataURL`. The session's resume
+// data is opaque, so the URL it belongs to is kept beside it.
+static NSString * const SQRLDownloaderResumeURLKey = @"URL";
+static NSString * const SQRLDownloaderResumeDataKey = @"resumeData";
+
+@implementation SQRLDownloadProgress
+
+- (instancetype)initWithBytesResumed:(int64_t)bytesResumed bytesReceived:(int64_t)bytesReceived bytesExpected:(int64_t)bytesExpected {
+	self = [super init];
+	if (self == nil) return nil;
+
+	_bytesResumed = bytesResumed;
+	_bytesReceived = bytesReceived;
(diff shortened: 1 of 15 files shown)


Squirrel/SQRLDownloader.m
```diff
@@ -0,0 +1,305 @@
...
+- (void)URLSession:(NSURLSession *)session task:(NSURLSessionTask *)task didCompleteWithError:(NSError *)error {
+	if (task != self.task) return;
+
+	if (error != nil) {
+		NSData *resumeData = error.userInfo[NSURLSessionDownloadTaskResumeData];
+		[self storeResumeData:resumeData];
+
+		// Resume data the session refused (stale temporary file, format it no
+		// longer reads): don't lose this attempt to it, start over once.
+		if (resumeData == nil && self.startedFromResumeData && !self.retriedFresh && !self.cancelled) {
```
MarshallOfSound: This only goes back to a fresh download when the session gives us *no* resume data back, but when the resumed request itself is what fails (a proxy/WAF that resets or blackholes `Range` requests, or a zip served with `Content-Encoding: gzip` which CFNetwork can't resume and fails with -1015) CFNetwork hands us fresh resume data at the same offset every time, so line 259 rewrites the plist and every check from here on (across relaunches too, `retriedFresh` is per instance) resumes and fails the exact same way and we never issue a plain GET again. Repro'd all three server behaviours, the only thing that eventually un-sticks it is the tmp cleaner deleting the partial ~3 days later. We should throw the stored resume data away when a resumed attempt fails without getting any further than it started (and fall into this fresh retry), or at minimum cap how many times we'll resume the same data.

Squirrel/SQRLDownloader.m
```diff
@@ -0,0 +1,305 @@
...
+		setNameWithFormat:@"%@ -downloadToURL: %@", self, destinationURL];
+}
+
+- (void)startTaskWithResumeData:(NSData *)resumeData {
+	dispatch_group_enter(self.completionGroup);
+
+	self.startedFromResumeData = (resumeData != nil);
+	self.bytesResumed = 0;
+	self.moveError = nil;
+	self.task = self.startedFromResumeData ? [self.session downloadTaskWithResumeData:resumeData] : [self.session downloadTaskWithRequest:self.request];
```
MarshallOfSound: Worth knowing that this replays the request archived inside the resume data rather than `self.request`, and it goes straight to the post-redirect URL. So for github release assets / presigned CDN redirects (i.e. what update.electronjs.org hands out) the resume hits the now expired signed URL and gets a 403, and a host injecting a rotating token via `requestForDownload` replays the old token. A 403/416/304 to the ranged request completes with `error == nil` so we don't hit the fresh retry below either, we just clear the plist and fail the check. Net effect on those setups is resume never actually works and every interruption costs two failed checks instead of the zero it costs on main. Not sure we can detect it up front 🤔 but when a task we started from resume data completes with a non-2xx we should probably just restart fresh in here rather than handing the 403 up, same shape as the `resumeData == nil` retry.

Also this is declared nonnull but if it ever did hand back nil we'd sit here forever with the command still executing (`-cancel` bails on `task == nil` too), a nil check that falls into the fresh path is cheap insurance.

Squirrel/SQRLDownloader.m
```diff
@@ -0,0 +1,305 @@
...
+	if (self.resumeDataURL == nil) return nil;
+
+	NSDictionary *stored = [NSDictionary dictionaryWithContentsOfURL:self.resumeDataURL];
+	NSData *resumeData = stored[SQRLDownloaderResumeDataKey];
+	if (![stored[SQRLDownloaderResumeURLKey] isEqual:self.request.URL.absoluteString] || ![resumeData isKindOfClass:NSData.class]) {
+		[self clearStoredResumeData];
+		return nil;
+	}
+
+	return resumeData;
```
MarshallOfSound, line 156: We should remove (or rename) the plist here when we hand the data to the session. As-is it stays on disk for the whole resumed attempt and CFNetwork appends into the same tmp file, so if the process dies mid resume (crash, or any Electron quit since the terminate hook doesn't fire there, see below) the next launch resumes from the stale offset, CFNetwork appends the range onto the now longer file and reports a 206 success. Repro'd it, a 4MB payload comes out as 5.2MB, we set the etag off it and then size/digest/ditto reject it and the rest of that launch gets 304s. Same root cause also lets two updaters (electron makes a new one on every `setFeedURL`) resume into the same tmp file at the same time, one gets the oversized file and the other ENOENTs on the move.

Squirrel/SQRLUpdater.m
```diff
@@ -516,17 +532,28 @@ - (RACSignal *)downloadBundleForUpdate:(SQRLUpdate *)update intoDirectory:(NSURL
 
 			[zipDownloadRequest setTimeoutInterval:SQURLUpdaterZipDownloadTimeoutSeconds];
 
-			return [[[NSURLConnection
-				rac_sendAsynchronousRequest:zipDownloadRequest]
-				reduceEach:^(NSURLResponse *response, NSData *bodyData) {
+			NSURL *zipOutputURL = [downloadDirectory URLByAppendingPathComponent:zipDownloadURL.lastPathComponent];
```
MarshallOfSound: This needs sanitizing now that the downloader `removeItemAtURL:`s the destination before moving into it. If the feed hands us a url whose last component is `..` (`/a/..`, `/%2E%2E` etc, `validateUpdateURL:` lets them all through) this resolves to the storage dir itself and `didFinishDownloadingToURL:` (which fires for a 404 as well) rm -rf's ShipItState.plist plus whatever update is already staged, and if ShipIt was already submitted for that staged update it then crash loops on the missing plist every 2s until logout because of the KeepAlive. On main the same url was just a failed `writeToURL:`. Bare host urls / `%00` in the name are broken too (those are pre-existing) so I'd just stop deriving this from the url entirely
```suggestion
			NSURL *zipOutputURL = [downloadDirectory URLByAppendingPathComponent:@"update.zip"];
```

Squirrel/SQRLUpdater.m
```diff
@@ -539,13 +566,75 @@ - (RACSignal *)downloadBundleForUpdate:(SQRLUpdate *)update intoDirectory:(NSURL
 						self.etag = httpResponse.allHeaderFields[@"ETag"];
```
MarshallOfSound, line 570: Ordering is pre-existing but this PR makes it bite a lot more: we store the etag before `verifyPackageAtURL:` runs, so a feed with a wrong `sha256` downloads and rejects once and then every later check in this process sends `If-None-Match`, gets a 304, reports no update and (via the `doCompleted` up in the command) flips state to `AwaitingRelaunch` with nothing staged. Fixing the feed json doesn't help because the zip's etag hasn't changed, only a relaunch does. Repro'd this end to end. We should only set this once the update has actually been prepared, or clear it in the `doError` in `downloadAndPrepareUpdate:`.

Squirrel/SQRLUpdate.m
```diff
@@ -65,6 +65,8 @@ + (NSDictionary *)JSONKeyPathsByPropertyKey {
 		@keypath(SQRLUpdate.new, releaseName): @"name",
 		@keypath(SQRLUpdate.new, releaseDate): @"pub_date",
 		@keypath(SQRLUpdate.new, updateURL): @"url",
+		@keypath(SQRLUpdate.new, packageDigest): @"sha256",
+		@keypath(SQRLUpdate.new, packageSize): @"size",
```
MarshallOfSound, line 69: Heads up that mapping these makes them fatal, Mantle's implicit class transformer rejects the whole model before `validatePackageSize:` is ever reached. So any feed that already happens to emit `size` as a string (or `sha256` as anything other than a string) for some other consumer goes from working to `SQRLUpdaterErrorInvalidJSON` on every check the moment the app picks up this Squirrel, and since updating is the thing that's broken there's no way to ship a client side fix. Both fields are optional so I think we should log + ignore malformed values rather than fail the model, or if we want them strict we should at least say so in the README.

Squirrel/SQRLUpdate.m
```diff
@@ -106,6 +108,27 @@ + (NSValueTransformer *)releaseDateJSONTransformer {
 
 #pragma mark NSKeyValueCoding
 
+- (BOOL)validatePackageDigest:(NSString **)stringPtr error:(NSError **)error {
```
MarshallOfSound: Can we validate the shape here as well (64 hex chars, and `size` > 0 in the one below)? Anything well typed but wrong (`""`, a `sha256:` prefix, base64, `size: 0`, the digest of the dmg instead of the zip) parses fine today and then we download the entire package on every launch just to throw it away, and after the first time the etag thing hides the error. Failing at parse is a lot cheaper and the error actually points at the feed. SQRLUpdateSpec currently asserts `ABCDEF` is a valid digest 😆

Squirrel/SQRLDownloader.m
```diff
@@ -0,0 +1,305 @@
...
+- (RACSignal *)downloadToURL:(NSURL *)destinationURL {
+	NSParameterAssert(destinationURL != nil);
+
+	return [[RACSignal
+		createSignal:^(id<RACSubscriber> subscriber) {
+			NSAssert(self.session == nil, @"%@ is one-shot; -downloadToURL: was subscribed to twice", self);
+
+			self.subscriber = subscriber;
+			self.destinationURL = destinationURL;
+			self.session = [NSURLSession sessionWithConfiguration:self.class.sessionConfiguration delegate:self delegateQueue:self.delegateQueue];
```
MarshallOfSound, line 194: This is a breaking change we should call out in the README: a session with its own configuration doesn't see anything registered via `+[NSURLProtocol registerClass:]` but the json request in SQRLUpdater still does, so a host intercepting its update traffic that way keeps passing the check and then fails every single download until it ships a build that sets `sessionConfiguration` (which is exactly what the OHHTTPStubs dance in SQRLUpdaterSpec is working around). Electron doesn't do this so I'm fine with the behaviour, just want it documented.

Squirrel/SQRLUpdater.m
```diff
@@ -198,6 +222,14 @@ - (id)initWithUpdateRequest:(NSURLRequest *)updateRequest requestForDownload:(SQ
 	}
 	_updateRequest = mutableUpdateRequest;
 	_updateClass = SQRLUpdate.class;
+	_downloadProgress = [[RACSubject subject] setNameWithFormat:@"%@ downloadProgress", self];
+
+	[[[NSNotificationCenter.defaultCenter
+		rac_addObserverForName:NSApplicationWillTerminateNotification object:nil]
+		takeUntil:self.rac_willDeallocSignal]
+		subscribeNext:^(id _) {
+			[SQRLDownloader cancelAllWritingResumeDataWithTimeout:SQRLUpdaterTerminationResumeDataTimeout];
```
MarshallOfSound, line 231: FYI this never fires under Electron, `ElectronApplication -terminate:` just calls `Browser::Quit()` and nothing in electron posts `NSApplicationWillTerminateNotification`. So for our main consumer resume only kicks in for network level failures and every quit mid download leaves a package sized `CFNetworkDownload_*.tmp` behind in the user's tmp dir. Doesn't break anything but the README line about this surviving a relaunch isn't true for electron, we should either note that or document `+cancelAllWritingResumeDataWithTimeout:` as the thing to call from your own quit path (it's public already) so electron can wire it up.

MarshallOfSound (commented):
Went through this pretty carefully and had most of the below repro'd against a local server rather than just reading it. The streaming side looks solid and nothing here lets a bad package actually get installed, the comments are mostly the resume state machine having a few ways to get stuck or do the wrong thing across checks/relaunches, plus the new feed fields being stricter than they look. The `..` url one, resume-fails-again, the plist not being consumed and the etag ordering are the ones I think we should sort out in this PR, the rest are docs/tests.

Squirrel/SQRLUpdater.m
```diff
@@ -539,13 +566,75 @@ - (RACSignal *)downloadBundleForUpdate:(SQRLUpdate *)update intoDirectory:(NSURL
...
+			if (update.packageDigest != nil) {
+				NSError *error;
+				NSFileHandle *handle = [NSFileHandle fileHandleForReadingFromURL:packageURL error:&error];
+				if (handle == nil) return [RACSignal error:error];
+
+				CC_SHA256_CTX context;
+				CC_SHA256_Init(&context);
+				for (;;) {
+					@autoreleasepool {
+						NSData *chunk = [handle readDataOfLength:1024 * 1024];
```
MarshallOfSound: `readDataOfLength:` raises on a read error (checked, EIO / the volume going away mid hash comes out as `NSFileHandleOperationException`) and we're inside the session's delegate callback here with nothing catching it, so a flaky disk during a background check takes the whole host app down. The `error:` variant is 10.15+ so probably just `read(2)` in a loop, or `@try` it and turn it into the same error. Line 554 has the same thing though that one's capped at 64k so I care less.

Squirrel/Squirrel.h
```diff
@@ -17,5 +17,6 @@ FOUNDATION_EXPORT const unsigned char SquirrelVersionString[];
 #import <Squirrel/NSBundle+SQRLVersionExtensions.h>
 #import <Squirrel/NSProcessInfo+SQRLVersionExtensions.h>
 #import <Squirrel/SQRLDownloadedUpdate.h>
+#import <Squirrel/SQRLDownloader.h>
```
MarshallOfSound, line 20: Note for whoever rolls this into electron, `build_add_gn_config.patch` enumerates the file list so SQRLDownloader.h/.m need adding there or the framework won't link and this import breaks auto_updater_mac.mm

SquirrelTests/SQRLUpdaterSpec.m
```diff
@@ -642,6 +651,37 @@ bool isRunningOnReadOnlyVolumeImp(id self, SEL _cmd)
...
+			return [request.URL.absoluteString isEqualToString:@"http://fake/download.zip"];
+		} withStubResponse:^(NSURLRequest *request) {
+			return [OHHTTPStubsResponse responseWithData:package statusCode:200 responseTime:0 headers:nil];
+		}];
+		[self addCleanupBlock:^{
+			[OHHTTPStubs removeRequestHandler:stubsCheck];
+			[OHHTTPStubs removeRequestHandler:stubsDownload];
+		}];
+
+		for (NSDictionary *mismatch in @[ @{ @"size": @(package.length + 1) }, @{ @"size": @(package.length), @"sha256": [@"" stringByPaddingToLength:64 withString:@"0" startingAtIndex:0] } ]) {
```
MarshallOfSound, line 677: Can we add the happy path here too (matching size + sha256 and the update actually gets prepared)? As written the second iteration can't tell a size pass from a size fail, and nothing in the suite proves a correct digest is accepted.

SquirrelTests/SQRLDownloaderSpec.m
```diff
@@ -0,0 +1,131 @@
...
+	expect([NSData dataWithContentsOfURL:destinationURL]).to(equal(payload));
+
+	SQRLDownloadProgress *last = progress.lastObject;
+	expect(@(last.bytesResumed)).to(equal(@0));
+	expect(@(last.bytesReceived)).to(equal(@(payloadLength)));
+	expect(@(last.bytesExpected)).to(equal(@(payloadLength)));
+	expect(@([NSFileManager.defaultManager fileExistsAtPath:resumeDataURL.path])).to(beFalsy());
+});
+
+it(@"should continue an interrupted transfer from where it stopped", ^{
```
MarshallOfSound, line 74: Would be good to cover the two nasty resume cases in here as well, a resume that fails again (server drops every ranged request, assert we eventually go back to a plain GET) and a resume from a plist that's stale relative to the tmp file. Those are the two I hit above and neither is exercised. Also the second `?drop=` request in this one only gets served in full because of the `dropped` set in TestServer.py which took me a minute to spot, worth a comment.

Squirrel/SQRLDownloader.m
```diff
@@ -0,0 +1,319 @@
...
+- (void)URLSession:(NSURLSession *)session task:(NSURLSessionTask *)task didCompleteWithError:(NSError *)error {
+	if (task != self.task) return;
+
+	// A resumed attempt that got no further than where it started says nothing
+	// about the payload: the session refused the data, the ranged request was
+	// reset or could not be decoded, or it was answered rather than served
+	// (403/416 from an expired signed URL, 304). Forget the resume data and
+	// ask from the start, once; never hand such an answer up as the download.
+	NSInteger statusCode = [task.response isKindOfClass:NSHTTPURLResponse.class] ? [(NSHTTPURLResponse *)task.response statusCode] : 200;
+	BOOL resumeWentNowhere = self.startedFromResumeData && !self.cancelled && (error != nil ? !self.receivedBytes : statusCode < 200 || statusCode > 299);
```
MarshallOfSound, line 271: This fixes the loop but takes the offline case with it: the plist is already gone from `takeStoredResumeData` and a resume that dies before any bytes (-1009 because we're offline, which is exactly the wifi dropped then slept case this feature is for, and the check on wake tends to fire before the network is back) counts as went nowhere so we never put it back, the fresh retry fails the same way and the whole partial is thrown away. Can we hang onto the resume data we took and re-store it if the fresh attempt also fails without receiving anything? That combo means the network is the problem rather than the ranged request, and if the fresh one gets anywhere at all we drop it like we do now.

Squirrel/SQRLDownloader.m
```diff
@@ -0,0 +1,319 @@
...
+	if (resumeWentNowhere) {
+		if (!self.retriedFresh) {
+			self.retriedFresh = YES;
+			dispatch_group_leave(self.completionGroup);
+			[self startTaskWithResumeData:nil];
+			return;
+		}
+		error = error ?: [NSError errorWithDomain:NSURLErrorDomain code:NSURLErrorCannotLoadFromNetwork userInfo:@{ NSURLErrorKey: self.request.URL }];
```
MarshallOfSound: Don't think we can get here, `resumeWentNowhere` needs `startedFromResumeData` and the fresh retry clears that, so `retriedFresh` is always NO by the time we're in this block. Can drop the else and the made up error

MarshallOfSound (commented):
Re-swept 898109d, everything from the first pass is sorted (re-ran the gzip / reset servers against the new nowhere check and neither gets a `didWriteData` before failing so both go fresh now, and the suite is green here). Two things on the new resume logic.

PR #328, opened 2026-08-11
feat: apply a binary delta from the running version when the server offers one

codebytere:
Stacked on #327 (base branch `stream-and-resume-downloads`); the diff here is only the delta work.

An update may now carry `delta`: a [Sparkle](https://sparkle-project.org) BinaryDelta patch (format 3 or 4) from one earlier `CFBundleVersion` to this release, with its URL, size and SHA-256:

```json
"delta": { "from_version": "412", "url": "https://…/412-to-413.delta", "sha256": "…", "size": 7340032 }
```

When `from_version` is the running application's, Squirrel downloads the patch instead of the ZIP (same streaming, resumable `SQRLDownloader`), checks size and digest before opening it, applies it to a copy of the running bundle inside the update directory, and sends the result through the existing code signing and `ElectronSquirrelPreventDowngrades` checks. If the delta is for another version, or downloading, verifying, applying or checking it fails in any way, the full ZIP is downloaded in the same check, so offering a delta can never cost an update. All four keys are required; ShipIt and installation are untouched.

The apply side is Sparkle's own, vendored unmodified under `Vendor/SparkleDelta` (MIT; format 3/4 reader and applier plus `bspatch`, ~2.9k lines, pinned to sparkle-project/Sparkle@cfa1190) with a small force-included config header standing in for Sparkle's xcconfig defines; the legacy XAR formats and the create side are left out, so patches are made with Sparkle's `BinaryDelta create old.app new.app out.delta` from the exact signed bundles that shipped. That keeps one implementation of the format rather than a Squirrel-flavoured second one. `objc_direct` is turned off for the vendored classes so the tests can link Sparkle's generator against them.

Sizing, for a sense of why: between two consecutive releases of a large Electron app (~230 MB zipped per arch, same Electron version) the format-4 LZMA delta is 9.5 MB, applies in a few seconds, and the result is byte-identical to the shipped bundle and passes `codesign --verify --deep --strict` and Gatekeeper.

Tests: `SQRLUpdaterSpec` builds real patches between the two TestApplication fixtures with Sparkle's generator (vendored into the test target only) and drives the out-of-process install twice, once from the delta alone (ZIP never requested) and once with a corrupted delta that must fall back to the ZIP, asserting on the fixture server's request log which files were fetched; `SQRLUpdateSpec` covers parsing and rejects a delta missing any key. 97 tests pass on macOS 26.5 / Xcode 26.5. README documents the key in both feed formats

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -16,3 +16,6 @@
 [submodule "Carthage/Checkouts/ReactiveObjC"]
 	path = Carthage/Checkouts/ReactiveObjC
 	url = https://github.com/ReactiveCocoa/ReactiveObjC.git
+[submodule "Carthage/Checkouts/Sparkle"]
+	path = Carthage/Checkouts/Sparkle
+	url = https://github.com/sparkle-project/Sparkle.git
--- a/README.md
+++ b/README.md
@@ -161,7 +161,13 @@ to the update request provided:
 	"notes": "Theses are some release notes innit",
 	"pub_date": "2013-09-18T12:29:53+01:00",
 	"sha256": "9f86d081884c7d659a2feaa0c55ad015a3bf4f1b2b0b822cd15d6c15b0f00a08",
-	"size": 104857600
+	"size": 104857600,
+	"delta": {
+		"from_version": "412",
+		"url": "https://mycompany.example.com/myapp/releases/412-to-myrelease.delta",
+		"sha256": "60303ae22b998861bce3b28f33eec1be758a213c86c93c076dbe9f558c11c752",
+		"size": 7340032
+	}
 }
 ```
 
@@ -193,6 +199,19 @@ path. Downloads run in their own `NSURLSession`, so an `NSURLProtocol`
 registered with `+registerClass:` sees the update check but not the download;
 set `SQRLDownloader.sessionConfiguration` (with its `protocolClasses`) to
 intercept both.
+
+"delta", if present, offers a binary patch from one earlier build to this
+release; it needs all four keys ("sha256" as 64 hex digits, "size" positive)
+and is logged and ignored otherwise. When "from_version" equals the
+running application's `CFBundleVersion`, Squirrel downloads the patch instead
+of the ZIP, checks its "size" and "sha256", applies it to a copy of the running
+application, and puts the result through the same code signing verification as
+an unpacked ZIP. If any of that fails, or "from_version" is anything else, it
+downloads the ZIP from "url" in the same check, so a server can always include
+the one delta it has for the version that asked. Patches are
+[Sparkle](https://sparkle-project.org) BinaryDelta files (format 3 or 4), made
+with Sparkle's `BinaryDelta create <old.app> <new.app> <patch>` from the exact
+signed bundles that were shipped.
 
 ## Update File JSON Format
 
@@ -250,6 +269,7 @@ file); a single entry is fine.
 | `updateTo.version` | — | Echoed into the `update-downloaded` event; conventionally the same as the outer `version`. |
 | `updateTo.name` / `notes` / `pub_date` | — | Surfaced to your app for display. `pub_date` must be ISO 8601 if present. |
 | `updateTo.sha256` / `size` | — | Digest (hex) and byte size of the `.zip`; a download that does not match is rejected. |
+| `updateTo.delta` | — | Optional `{from_version, url, sha256, size}` binary patch from one earlier `CFBundleVersion`; tried first when it matches the running app, with the `.zip` as fallback. |
 
 Point the updater directly at this file's URL — there's no required filename.
 
--- a/README.md
+++ b/README.md
@@ -45,6 +45,13 @@ repository (under `Carthage/Checkouts/`, a directory name kept from when they
 were fetched with Carthage); `git submodule update --init` or `script/bootstrap`
 checks them out.
 
+Binary delta support compiles Sparkle's BinaryDelta sources and the bsdiff it
+vendors straight out of a third submodule, `Carthage/Checkouts/Sparkle`, pinned
+to a Sparkle release tag (currently 2.9.5); nothing of Sparkle is linked as a
+framework and applications need not ship it. To move the pin, check out the new
+tag in the submodule, build, run the tests, and commit the submodule change; the
+files involved are listed under the SparkleDelta group in the Xcode project.
+
 If your application is already using ReactiveObjC, ensure it is using the same
 version as Squirrel.
 
@@ -210,11 +217,22 @@ running application's `CFBundleVersion`, Squirrel downloads the patch instead
 of the ZIP, checks its "size" and "sha256", applies it to a copy of the running
 application, and puts the result through the same code signing verification as
 an unpacked ZIP. If any of that fails, or "from_version" is anything else, it
-downloads the ZIP from "url" in the same check, so a server can always include
-the one delta it has for the version that asked. Patches are
-[Sparkle](https://sparkle-project.org) BinaryDelta files (format 3 or 4), made
-with Sparkle's `BinaryDelta create <old.app> <new.app> <patch>` from the exact
-signed bundles that were shipped.
+downloads the ZIP from "url" in the same check (its `downloadProgress` starting
+again from zero), so a server can always include the one delta it has for the
+version that asked. A delta that has been applied and staged is not fetched
+again by later checks in the same process.
+
+Patches are [Sparkle](https://sparkle-project.org) BinaryDelta files (format 3
+or 4, any `--compression` except `bzip2`), made with Sparkle's
+`BinaryDelta create <old.app> <new.app> <patch>`. A patch only applies to a
+byte-identical copy of `<old.app>`, file modes included, and ShipIt clears the
+group and other write bits of everything it installs; strip them from the app
+before signing it (`chmod -R go-w MyApp.app`) so the shipped bundle, the
+installed bundle and the trees the patch was made from all agree, otherwise
+the patch applies once to a fresh install and never again. Files the patch adds
+or rewrites are created with decomposed (NFD) names, so a non-ASCII file name
+outside an archive such as `app.asar` can fail code signing verification and
+cost a fallback to the ZIP. "from_version" may be a JSON string or number.
 
 ## Update File JSON Format
 
--- a/README.md
+++ b/README.md
@@ -220,7 +220,8 @@ an unpacked ZIP. If any of that fails, or "from_version" is anything else, it
 downloads the ZIP from "url" in the same check (its `downloadProgress` starting
 again from zero), so a server can always include the one delta it has for the
 version that asked. A delta that has been applied and staged is not fetched
-again by later checks in the same process.
+again by later checks in the same process, and neither is one that downloaded
+intact but would not apply or verify.
 
 Patches are [Sparkle](https://sparkle-project.org) BinaryDelta files (format 3
 or 4, any `--compression` except `bzip2`), made with Sparkle's
--- a/Squirrel/SQRLUpdateDelta.m
+++ b/Squirrel/SQRLUpdateDelta.m
@@ -19,7 +19,7 @@
 	if (self.fromVersion.length == 0) missing = @"from_version";
 	else if (self.deltaURL.scheme == nil || self.deltaURL.path == nil) missing = @"url";
 	else if (self.digest.length != 64 || [self.digest rangeOfCharacterFromSet:notHex].location != NSNotFound) missing = @"sha256";
-	else if (self.size.longLongValue <= 0) missing = @"size";
+	else if (![self.size isKindOfClass:NSNumber.class] || CFGetTypeID((__bridge CFTypeRef)self.size) == CFBooleanGetTypeID() || CFNumberIsFloatType((__bridge CFNumberRef)self.size) || self.size.longLongValue <= 0) missing = @"size";
 
 	if (missing != nil) {
 		if (error != NULL) {
--- a/Squirrel/SQRLUpdateDelta.m
+++ b/Squirrel/SQRLUpdateDelta.m
@@ -1,0 +1,58 @@
+//
+//  SQRLUpdateDelta.m
+//  Squirrel
+//
+
+#import "SQRLUpdateDelta.h"
+#import <ReactiveObjC/EXTKeyPathCoding.h>
+
+@implementation SQRLUpdateDelta
+
+#pragma mark Lifecycle
+
+- (id)initWithDictionary:(NSDictionary *)dictionary error:(NSError **)error {
+	self = [super initWithDictionary:dictionary error:error];
+	if (self == nil) return nil;
+
+	NSCharacterSet *notHex = [NSCharacterSet characterSetWithCharactersInString:@"0123456789abcdef"].invertedSet;
+	NSString *missing = nil;
+	if (self.fromVersion.length == 0) missing = @"from_version";
+	else if (self.deltaURL.scheme == nil || self.deltaURL.path == nil) missing = @"url";
+	else if (self.digest.length != 64 || [self.digest rangeOfCharacterFromSet:notHex].location != NSNotFound) missing = @"sha256";
+	else if (self.size.longLongValue <= 0) missing = @"size";
+
+	if (missing != nil) {
+		if (error != NULL) {
+			*error = [NSError errorWithDomain:NSCocoaErrorDomain code:NSKeyValueValidationError userInfo:@{
+				NSLocalizedDescriptionKey: NSLocalizedString(@"Validation failed", nil),
+				NSLocalizedRecoverySuggestionErrorKey: [NSString stringWithFormat:NSLocalizedString(@"A delta update needs a valid \"%@\".", nil), missing],
+			}];
+		}
+		return nil;
+	}
+
+	return self;
+}
+
+#pragma mark MTLJSONSerializing
+
++ (NSDictionary *)JSONKeyPathsByPropertyKey {
+	return @{
+		@keypath(SQRLUpdateDelta.new, fromVersion): @"from_version",
+		@keypath(SQRLUpdateDelta.new, deltaURL): @"url",
+		@keypath(SQRLUpdateDelta.new, digest): @"sha256",
+		@keypath(SQRLUpdateDelta.new, size): @"size",
+	};
+}
+
++ (NSValueTransformer *)deltaURLJSONTransformer {
+	return [NSValueTransformer valueTransformerForName:MTLURLValueTransformerName];
+}
+
++ (NSValueTransformer *)digestJSONTransformer {
+	return [MTLValueTransformer transformerUsingForwardBlock:^ id (NSString *digest, BOOL *success, NSError **error) {
+		return [digest isKindOfClass:NSString.class] ? digest.lowercaseString : nil;
+	}];
+}
+
+@end
--- a/Squirrel/SQRLUpdater.m
+++ b/Squirrel/SQRLUpdater.m
@@ -17,6 +17,8 @@
 #import "SQRLDownloader.h"
 #import "SQRLShipItLauncher.h"
 #import "SQRLUpdate.h"
+#import "SQRLUpdateDelta.h"
+#import "SUBinaryDeltaApply.h"
 #import "SQRLZipArchiver.h"
 #import "SQRLShipItRequest.h"
 #import <ReactiveObjC/EXTScope.h>
@@ -117,12 +119,30 @@ BOOL isVersionStandard(NSString* version) {
 // errors, on a background thread.
 - (RACSignal *)downloadBundleForUpdate:(SQRLUpdate *)update intoDirectory:(NSURL *)downloadDirectory;
 
-// Checks a downloaded package against the size and digest its update declares.
+// Checks a downloaded file against a declared size and SHA-256 digest, deleting
+// it on mismatch.
 //
 // Returns a signal which completes, or errors with
 // `SQRLUpdaterErrorInvalidUpdatePackage`, on a background thread. Completes
-// immediately when the update declares neither.
-- (RACSignal *)verifyPackageAtURL:(NSURL *)packageURL forUpdate:(SQRLUpdate *)update;
+// immediately when neither is declared.
+- (RACSignal *)verifyFileAtURL:(NSURL *)fileURL size:(NSNumber *)size digest:(NSString *)digest;
+
+// Produces the verified update in `downloadDirectory`: from `update.delta`
+// applied to the running application when that delta targets the running
+// version, and otherwise, or if anything about the delta fails, from the
+// full archive at `update.updateURL`.
+//
+// Returns a signal which sends a `SQRLDownloadedUpdate` (or nil when a
+// conditional GET says the archive was already downloaded) then completes, or
+// errors, on a background thread.
+- (RACSignal *)downloadedUpdateForUpdate:(SQRLUpdate *)update intoDirectory:(NSURL *)downloadDirectory;
+
+// Downloads `delta`, verifies it, and applies it to a copy of the running
+// application inside `downloadDirectory`.
+//
+// Returns a signal which sends the patched `NSBundle` then completes, or
+// errors, on a background thread.
+- (RACSignal *)bundleByApplyingDelta:(SQRLUpdateDelta *)delta intoDirectory:(NSURL *)downloadDirectory;
 
 // Creates a unique directory in which to save the update bundle, for later use
 // by ShipIt.
@@ -140,17 +160,14 @@ BOOL isVersionStandard(NSString* version) {
 // errors.
 - (RACSignal *)updateBundleMatchingCurrentApplicationInDirectory:(NSURL *)directory;
 
-// Validates the code signature of the given update bundle, then prepares it for
-// installation.
+// Validates the code signature (and, with ElectronSquirrelPreventDowngrades,
+// the version) of the given update bundle.
 //
-// Upon success, the update will be automatically installed after the
-// application terminates.
-//
-// update - Describes the update to verify and prepare. This must not be nil.
+// update - Describes the update to verify. This must not be nil.
 //
 // Returns a signal which sends a `SQRLDownloadedUpdate` then completes, or
 // errors, on a background thread.
-- (RACSignal *)verifyAndPrepareUpdate:(SQRLUpdate *)update fromBundle:(NSBundle *)updateBundle;
+- (RACSignal *)verifyUpdate:(SQRLUpdate *)update fromBundle:(NSBundle *)updateBundle;
 
 // Prepares the given update for installation.
 //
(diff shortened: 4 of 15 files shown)


MarshallOfSound (commented):
Went through this the same way as #327 (read it all, then had most of it repro'd against the built framework / real ShipIt). The apply itself holds up, I couldn't get anything bad installed, Sparkle's hashes plus our codesign check do their job. The comments are mostly about the delta path quietly degrading into delta + zip for everyone (modes, NFD, resume) plus one actual regression with locked files. The re-apply on every check one and the locked file one I'd want sorted in this PR, the rest is docs / roll notes. A few small things not worth their own thread: PreventDowngrades failures now cost a delta + clone + apply before the zip produces the exact same error, Sparkle leaks its temp diff in $TMPDIR when extraction fails (upstream), the clone carries the old root's quarantine xattr forward and ShipIt's clearQuarantine skips the root (checked, doesn't translocate, it just differs from a zip install), and the PR description still says this is vendored under Vendor/ when it's a submodule now.

Squirrel/SQRLUpdater.m
```diff
@@ -503,6 +522,104 @@ - (RACSignal *)downloadAndPrepareUpdate:(SQRLUpdate *)update {
...
+
+	NSURL *runningURL = NSRunningApplication.currentApplication.bundleURL;
+	NSString *runningVersion = (runningURL == nil ? nil : [NSBundle bundleWithURL:runningURL].sqrl_bundleVersion);
+	if (![delta.fromVersion isEqual:runningVersion]) {
+		NSLog(@"Delta update is from %@ but %@ is running, downloading the full update instead", delta.fromVersion, runningVersion);
+		return fullUpdate;
+	}
+
+	return [[[[self
+		bundleByApplyingDelta:delta intoDirectory:downloadDirectory]
```
MarshallOfSound, line 567: Nothing short circuits this once we've already staged this exact update. The zip path gets a 304 off `self.etag` on every later check and does nothing, but `from_version` still matches the running app, so every check from here until relaunch re-downloads the delta (no If-None-Match), re-clones and re-patches the running app, re-runs the full codesign check, stages a fresh update.XXXXXXX and re-emits the update. Repro'd with the built TestApplication, 3 checks = 3 delta downloads, 3 staged dirs, 3 emissions. In electron that's an `update-downloaded` event per check so update-electron-app pops its restart dialog every 10 minutes 😬 and it makes the quitAndInstall race real too, a check that finishes mid quit rewrites the plist with launchAfterInstallation NO and the app doesn't come back. We should skip the delta path when we're already AwaitingRelaunch for this update, remembering the digest of the delta we staged is probably the simplest way.

Squirrel/SQRLUpdater.m
```diff
@@ -503,6 +522,104 @@ - (RACSignal *)downloadAndPrepareUpdate:(SQRLUpdate *)update {
...
+		flattenMap:^(NSBundle *updateBundle) {
+			return [self verifyUpdate:update fromBundle:updateBundle];
+		}]
+		catch:^(NSError *error) {
+			NSLog(@"Delta update from %@ could not be used (%@), downloading the full update instead", delta.fromVersion, error.sqrl_verboseDescription);
+
+			NSFileManager *fileManager = [[NSFileManager alloc] init];
+			for (NSURL *leftover in [fileManager contentsOfDirectoryAtURL:downloadDirectory includingPropertiesForKeys:nil options:0 error:NULL]) {
```
MarshallOfSound: Falling back into the same directory the delta just dirtied is what turns a couple of odd but real setups into never-updates-again. Sparkle's before hash covers modes but not st_flags, so a Finder locked (uchg) file inside the running app passes and `copyItem` faithfully copies the flag onto the clone, at which point this loop can't delete it (error is swallowed), ditto then aborts extracting the zip over the top of it and the check fails, and cleanUp / prune can't remove the dir either. Repro'd against the real build: every automatic check leaves a full update.zip plus the locked skeleton behind in Caches, and if the delta happens not to touch the locked file we get further and ShipIt fails instead (removexattr gets EPERM on the immutable copy, 3 attempts then it gives up). Same locked file on the zip only path installs fine today so this one's a regression. We should fall back into a fresh update dir, and probably chflags the clone before trying to remove it.

Squirrel/SQRLUpdater.m
```diff
@@ -503,6 +522,104 @@ - (RACSignal *)downloadAndPrepareUpdate:(SQRLUpdate *)update {
...
+
+	return [[RACSignal
+		defer:^{
+			NSMutableURLRequest *request = [self.requestForDownload(delta.deltaURL) mutableCopy];
+			[request setValue:@"application/octet-stream" forHTTPHeaderField:@"Accept"];
+			[request setTimeoutInterval:SQURLUpdaterZipDownloadTimeoutSeconds];
+
+			// A fixed name: the URL is the server's, the directory is ours.
+			NSURL *deltaOutputURL = [downloadDirectory URLByAppendingPathComponent:@"update.delta"];
+			NSURL *resumeDataURL = [downloadDirectory.URLByDeletingLastPathComponent URLByAppendingPathComponent:SQRLUpdaterResumeDataFileName];
```
MarshallOfSound: This is the same file the zip path uses and `takeStoredResumeData` deletes it whenever the URL doesn't match, so with a delta in the feed the resume stuff from #327 mostly stops working. Delta download gets interrupted, we fall back to the zip in the same check and the zip downloader deletes the delta's resume data, so delta resume never works, and if the delta is failing persistently for someone (404 / not on the CDN yet / the hash issues below) every check runs the delta downloader first which deletes the zip's resume data and the big download starts from zero every time. Repro'd both directions through the real `downloadedUpdateForUpdate:`. Just needs its own name (constant next to the other one plz)
```suggestion
			NSURL *resumeDataURL = [downloadDirectory.URLByDeletingLastPathComponent URLByAppendingPathComponent:@"delta.resumedata"];
```

Squirrel/SQRLUpdater.m
```diff
@@ -503,6 +522,104 @@ - (RACSignal *)downloadAndPrepareUpdate:(SQRLUpdate *)update {
...
+
+					NSURL *sourceURL = NSRunningApplication.currentApplication.bundleURL.URLByResolvingSymlinksInPath;
+					NSURL *patchedURL = [downloadDirectory URLByAppendingPathComponent:sourceURL.lastPathComponent];
+					return [[[self
+						verifyFileAtURL:deltaFileURL size:delta.size digest:delta.digest]
+						then:^{
+							return [RACSignal startLazilyWithScheduler:[RACScheduler schedulerWithPriority:RACSchedulerPriorityBackground] block:^(id<RACSubscriber> subscriber) {
+								NSLog(@"Applying delta %@ to %@", deltaFileURL.lastPathComponent, sourceURL.path);
+								NSError *error = nil;
+								BOOL applied = applyBinaryDelta(sourceURL.path, patchedURL.path, deltaFileURL.path, NO, ^(double progress){}, &error);
```
MarshallOfSound, line 653: Bigger problem with pointing this at the installed app: ShipIt strips the g/o write bits off every file it installs (`takeOwnershipOfFileSecurity:` in SQRLInstaller) and Sparkle's before hash includes the mode of every file and dir, so if the shipped bundle has a single 0664 / 0775 entry in it (umask 002 CI, stuff pulled out of a tarball into app.asar.unpacked, stock electron itself is clean) the delta only ever applies to a copy that has never been through ShipIt. Repro'd with the actual ShipIt binary, hand installed copy applies, ShipIt installed copy fails the before hash in both format 3 and 4, and the output of a successful delta still carries the bits so the next one fails as well. So for an affected publisher it works exactly once while they're testing it and then it's delta + zip for the whole fleet on every release forever, with an NSLog as the only signal. At minimum the README needs to tell people to `chmod -R go-w` before signing, but given we control both ends I'd rather we either stop stripping bits we don't need to or normalise on our side.

Squirrel/SQRLUpdater.m
```diff
@@ -503,6 +522,104 @@ - (RACSignal *)downloadAndPrepareUpdate:(SQRLUpdate *)update {
...
+	NSString *runningVersion = (runningURL == nil ? nil : [NSBundle bundleWithURL:runningURL].sqrl_bundleVersion);
+	if (![delta.fromVersion isEqual:runningVersion]) {
+		NSLog(@"Delta update is from %@ but %@ is running, downloading the full update instead", delta.fromVersion, runningVersion);
+		return fullUpdate;
+	}
+
+	return [[[[self
+		bundleByApplyingDelta:delta intoDirectory:downloadDirectory]
+		flattenMap:^(NSBundle *updateBundle) {
+			return [self verifyUpdate:update fromBundle:updateBundle];
```
MarshallOfSound: One more way this quietly ends up as delta + zip for everyone: Sparkle writes every path through `fileSystemRepresentation`, so anything the applier creates (added files, modified files under 4k, clones, new dirs) lands on disk NFD while everything copyTree brought over from the ditto extracted install is NFC. Sparkle's own hash goes through the same normalisation so the apply succeeds and then this fails with errSecCSBadResource because CodeResources sealed the NFC name (repro'd with SQRLCodeSignature's actual flags, the file shows up as both added and missing). Bigger files get bspatched in place and keep their name so it's a subset, and most electron apps only have ASCII names outside the asar, but it's upstream behaviour we're inheriting so worth a README note at least and probably an issue on Sparkle.

Squirrel/SQRLUpdater.m
```diff
@@ -503,6 +522,104 @@ - (RACSignal *)downloadAndPrepareUpdate:(SQRLUpdate *)update {
...
+						}
+					}
+
+					NSURL *sourceURL = NSRunningApplication.currentApplication.bundleURL.URLByResolvingSymlinksInPath;
+					NSURL *patchedURL = [downloadDirectory URLByAppendingPathComponent:sourceURL.lastPathComponent];
+					return [[[self
+						verifyFileAtURL:deltaFileURL size:delta.size digest:delta.digest]
+						then:^{
+							return [RACSignal startLazilyWithScheduler:[RACScheduler schedulerWithPriority:RACSchedulerPriorityBackground] block:^(id<RACSubscriber> subscriber) {
+								NSLog(@"Applying delta %@ to %@", deltaFileURL.lastPathComponent, sourceURL.path);
```
MarshallOfSound, line 651: FYI on footprint, bspatch reads the whole old file in and mallocs the whole new file per patched file, so for an electron app this is roughly 2x the framework binary inside the app process (measured ~390MB single arch, ~780MB universal) and libmalloc doesn't really hand it back afterwards, plus it's per check until the comment above is sorted. #327 just got rid of a 1x buffer 😆 Sparkle does this bit in a helper process and we already ship one, so worth thinking about doing the apply out of process at some point, not blocking on it.

Squirrel/SQRLUpdater.m
```diff
@@ -17,6 +17,8 @@
 #import "SQRLDownloader.h"
 #import "SQRLShipItLauncher.h"
 #import "SQRLUpdate.h"
+#import "SQRLUpdateDelta.h"
+#import "SUBinaryDeltaApply.h"
```
MarshallOfSound, line 22: Note for whoever rolls this into electron (not for this PR): gclient doesn't init submodules so Sparkle needs a DEPS entry like Mantle / ReactiveObjC, and the GN patch needs SQRLUpdateDelta.h (plus SQRLDownloader.h from #327) in headers, the new .m / .c files, the include dirs, `-lz -lcompression` and a licenses.py entry. This quoted import only resolves through the xcode headermap today so it'll be the first thing to fall over.

.gitmodules
```diff
@@ -16,3 +16,6 @@
 [submodule "Carthage/Checkouts/ReactiveObjC"]
 	path = Carthage/Checkouts/ReactiveObjC
 	url = https://github.com/ReactiveCocoa/ReactiveObjC.git
+[submodule "Carthage/Checkouts/Sparkle"]
+	path = Carthage/Checkouts/Sparkle
+	url = https://github.com/sparkle-project/Sparkle.git
```
MarshallOfSound, line 21: This is the tip of Sparkle's 2.x branch rather than a tag (the files we actually build happen to be identical to 2.9.5 but nothing says so). Can we pin to the release tag and put a line in the README about how to bump it? Also we're now shipping Sparkle (MIT) and bsdiff (BSD) code inside the framework itself, so their notices should go in LICENSE.

Squirrel/SparkleDelta/SparkleDeltaConfig.h
```diff
@@ -0,0 +1,16 @@
+// Build configuration for the Sparkle BinaryDelta sources Squirrel compiles
+// out of the Sparkle submodule (Carthage/Checkouts/Sparkle). Sparkle sets these
+// from its xcconfigs; Squirrel force-includes this header instead.
+
+// Only Sparkle's own container format (delta major versions 3 and 4) is read;
+// the XAR-based formats 1 and 2 are not built.
+#define SPARKLE_BUILD_LEGACY_DELTA_SUPPORT 0
+// LZMA/LZFSE/LZ4/ZLIB via libcompression only.
+#define SPARKLE_BUILD_BZIP2_DELTA_SUPPORT 0
```
MarshallOfSound: Comment above isn't quite right, bspatch still goes through BZ2_* for the per file patches which is why `-lbz2` is still needed. Refusing bzip2 containers is fine (default is lzma and I checked it errors cleanly), just worth a README line that `--compression bzip2` will never apply.

README.md
```diff
@@ -194,6 +200,19 @@ registered with `+registerClass:` sees the update check but not the download;
...
+and is logged and ignored otherwise. When "from_version" equals the
+running application's `CFBundleVersion`, Squirrel downloads the patch instead
+of the ZIP, checks its "size" and "sha256", applies it to a copy of the running
+application, and puts the result through the same code signing verification as
+an unpacked ZIP. If any of that fails, or "from_version" is anything else, it
+downloads the ZIP from "url" in the same check, so a server can always include
+the one delta it has for the version that asked. Patches are
+[Sparkle](https://sparkle-project.org) BinaryDelta files (format 3 or 4), made
+with Sparkle's `BinaryDelta create <old.app> <new.app> <patch>` from the exact
+signed bundles that were shipped.
```
MarshallOfSound: Given the threads on SQRLUpdater.m this bit should probably also say: strip g/o write bits before signing (or however we end up handling that), non ASCII names outside the asar can make a delta fall back, don't use `--compression bzip2`, `from_version` has to be a JSON string (a number gets silently dropped), and that `downloadProgress` starts again from 0 for the zip when the delta fails.

Squirrel/SQRLUpdateDelta.m
```diff
@@ -0,0 +1,58 @@
...
+- (id)initWithDictionary:(NSDictionary *)dictionary error:(NSError **)error {
+	self = [super initWithDictionary:dictionary error:error];
+	if (self == nil) return nil;
+
+	NSCharacterSet *notHex = [NSCharacterSet characterSetWithCharactersInString:@"0123456789abcdef"].invertedSet;
+	NSString *missing = nil;
+	if (self.fromVersion.length == 0) missing = @"from_version";
+	else if (self.deltaURL.scheme == nil || self.deltaURL.path == nil) missing = @"url";
+	else if (self.digest.length != 64 || [self.digest rangeOfCharacterFromSet:notHex].location != NSNotFound) missing = @"sha256";
+	else if (self.size.longLongValue <= 0) missing = @"size";
```
MarshallOfSound: Two small ones on the parsing: a JSON number for `from_version` never gets here (Mantle drops the whole delta before this runs) and CFBundleVersions are numbers more often than not, so we should probably stringify rather than drop. And `true` / `1.5` get through this size check and then fail the isEqual: in verify after we've downloaded the whole delta, on every check, so this should require an integer.

SquirrelTests/SQRLUpdaterSpec.m
```diff
@@ -424,6 +426,76 @@ bool isRunningOnReadOnlyVolumeImp(id self, SEL _cmd)
...
+			for (NSString *line in [[NSString stringWithContentsOfURL:requestLogURL encoding:NSUTF8StringEncoding error:NULL] componentsSeparatedByString:@"\n"]) {
+				if (line.length > 0) [paths addObject:[line componentsSeparatedByString:@" "][1]];
+			}
+			return paths;
+		};
+
+		SQRLUpdateDelta * (^servedDelta)(BOOL) = ^(BOOL intact) {
+			NSURL *deltaURL = [serverDirectoryURL URLByAppendingPathComponent:@"update.delta"];
+			NSError *error = nil;
+			BOOL created = createBinaryDelta(self.testApplicationURL.path, updateURL.path, deltaURL.path, SUBinaryDeltaMajorVersion4, SPUDeltaCompressionModeDefault, 0, NO, &error);
```
MarshallOfSound: Both of these build the delta from the exact tree the app is running out of, which is why neither the mode stripping nor the NFD thing show up. We should have one that installs v1 via ShipIt first and then applies a delta built from the pristine bundles, that's the case that actually happens in the wild. Also missing: from_version mismatch (assert the delta isn't even requested), digest mismatch, a mid apply failure that leaves a partial bundle behind, and a second check after staging.

Squirrel/SQRLUpdater.m
codebytere: 9bd1c83 remembers the digest of the delta it applied and staged (cleared wherever the ETag is) and answers a later check offering the same delta the way a 304 answers the zip path, so nothing is re-downloaded, re-patched or re-emitted until relaunch. New spec runs two checks and asserts one delta request.

Squirrel/SQRLUpdater.m
codebytere: 9bd1c83 falls back into a fresh update directory and clears file flags on the dirtied one before removing it (removal failure is logged, not fatal). New spec corrupts the payload past the header so the apply fails after writing, and still installs from the zip.

Squirrel/SQRLUpdater.m
codebytere: Done, `delta.resumedata` as its own constant next to the other.

Squirrel/SQRLUpdater.m
codebytere: Agree it's the sharp edge. The before-hash is over the publisher's tree, so the apply side can't repair a delta made from g+w files; the one normal form both ends can share is what ShipIt produces. README now says to `chmod -R go-w` before signing so shipped, installed and delta trees agree, and calls out the applies-once-then-never failure mode explicitly. I'd rather keep takeOwnership stripping g/o write than reopen installed apps to group writes. New spec installs v1 through ShipIt first and then applies a delta made from the as-built bundles.

Squirrel/SQRLUpdater.m
codebytere: README note added (files the patch creates land NFD, so non-ASCII names outside the asar can fail signing and cost a fallback). Happy to raise it with Sparkle separately.

Squirrel/SQRLUpdater.m
codebytere: Ack, and agreed on doing the apply out of process eventually; with the staged-once fix it's at least once per update rather than once per check. Leaving for a follow-up.

Squirrel/SQRLUpdater.m
codebytere: Noted for the roll: DEPS entry for Sparkle, GN headers/sources incl. SQRLUpdateDelta.h and SQRLDownloader.h, include dirs, `-lz -lcompression -lbz2`, licenses.py.

.gitmodules
codebytere: Pinned to the 2.9.5 tag (the sources we compile are identical to the previous sha), README's Dependencies section says how to bump it, and LICENSE now carries Sparkle's MIT text plus the bsdiff and sais-lite notices.

Squirrel/SparkleDelta/SparkleDeltaConfig.h
codebytere: Fixed the comment (containers via libcompression only, per-file patches still bzip2 hence libbz2) and the README says `--compression bzip2` containers won't apply.

README.md
codebytere: All five are in the delta section now: go-w before signing, NFD names, no bzip2 containers, from_version as string or number (numbers are now accepted, see the parsing thread), progress restarting from zero on fallback.

Squirrel/SQRLUpdateDelta.m
codebytere: 9bd1c83 stringifies a numeric from_version before Mantle sees it, and size must be a positive integer (bools and floats are dropped with the rest). Unit spec updated both ways.

SquirrelTests/SQRLUpdaterSpec.m
codebytere: Added: ShipIt-installed v1 then a delta made from the shipped bundles; from_version mismatch (asserts only /full.zip is requested); digest mismatch; header corruption; mid-apply corruption; two checks after staging (one delta request). 105 green locally on Xcode 26.

Squirrel/SQRLUpdater.m
```diff
@@ -482,27 +509,150 @@ - (RACSignal *)downloadAndPrepareUpdate:(SQRLUpdate *)update {
...
+
+	// Already applied and staged from this delta: nothing to download, the
+	// same answer a 304 gives the ZIP path.
+	if ([delta.digest isEqual:self.appliedDeltaDigest]) return [RACSignal return:nil];
+
+	return [[[[self
+		bundleByApplyingDelta:delta intoDirectory:downloadDirectory]
+		flattenMap:^(NSBundle *updateBundle) {
+			return [[self verifyUpdate:update fromBundle:updateBundle] doCompleted:^{
+				self.appliedDeltaDigest = delta.digest;
```
MarshallOfSound: The other half of the locked file case is still there on this path. If the delta doesn't touch the locked file the before hash passes, the apply and this verify both succeed, and the bundle we stage still has uchg on it (copyTree brings the flag across), so ShipIt then dies in clearQuarantineForDirectory (removexattr gives EPERM rather than ENOATTR on an immutable file), retries 3 times and gives up. That's the case I actually repro'd end to end with ShipIt on the previous push and nothing changed for it, and the zip path installs the same app fine, so it's still a regression, now once per launch instead of once per check because of the memo. Simplest fix is probably to run the same lchflags loop from `removeDirectoryClearingFlagsAtURL:` over the patched bundle before we verify it, then the helper does double duty.

Squirrel/SQRLUpdater.m
```diff
@@ -482,27 +509,150 @@ - (RACSignal *)downloadAndPrepareUpdate:(SQRLUpdate *)update {
...
+	NSURL *runningURL = NSRunningApplication.currentApplication.bundleURL;
+	NSString *runningVersion = (runningURL == nil ? nil : [NSBundle bundleWithURL:runningURL].sqrl_bundleVersion);
+	if (![delta.fromVersion isEqual:runningVersion]) {
+		NSLog(@"Delta update is from %@ but %@ is running, downloading the full update instead", delta.fromVersion, runningVersion);
+		return fullUpdateIntoDirectory(downloadDirectory);
+	}
+
+	// Already applied and staged from this delta: nothing to download, the
+	// same answer a 304 gives the ZIP path.
+	if ([delta.digest isEqual:self.appliedDeltaDigest]) return [RACSignal return:nil];
```
MarshallOfSound, line 563: This only remembers the ones that worked, so a delta that's failing deterministically for someone (the go-w / NFD cases) still gets downloaded and cloned + applied on every check until relaunch. Not a big deal now that the success case is memoised, but we could remember apply / verify failures too (not download ones) and it becomes once per process as well 🤷

Squirrel/SQRLUpdater.m
```diff
@@ -482,27 +509,150 @@ - (RACSignal *)downloadAndPrepareUpdate:(SQRLUpdate *)update {
...
+			}];
+		}]
+		catch:^(NSError *error) {
+			NSLog(@"Delta update from %@ could not be used (%@), downloading the full update instead", delta.fromVersion, error.sqrl_verboseDescription);
+
+			// Whatever the apply left behind may not be removable (it copies
+			// file flags such as uchg from the running app), so the ZIP gets a
+			// directory of its own rather than this one emptied.
+			[self removeDirectoryClearingFlagsAtURL:downloadDirectory];
+			return [[self uniqueTemporaryDirectoryForUpdate] flattenMap:^(NSURL *freshDirectory) {
```
MarshallOfSound, line 594: Nit: `cleanUp()` up in `downloadAndPrepareUpdate:` still points at the original directory, which we've just removed here, so every fallback now logs an "Error removing temporary download directory" and if the zip side then errors or 304s the fresh dir hangs around until the next check prunes it. Bounded so not fussed, but having cleanUp track whichever dir we ended up using would tidy both up.

SquirrelTests/SQRLUpdaterSpec.m
```diff
@@ -424,6 +426,128 @@ bool isRunningOnReadOnlyVolumeImp(id self, SEL _cmd)
...
+			updateWithDelta(deltaWith(servedDelta(self.testApplicationURL, intact), @{ @"sha256": [@"" stringByPaddingToLength:64 withString:@"0" startingAtIndex:0] }), nil);
+			expect(requestedPaths()).to(equal(@[ @"/update.delta", @"/full.zip" ]));
+		});
+
+		it(@"should fall back to the full update when the delta does not apply", ^{
+			updateWithDelta(servedDelta(self.testApplicationURL, NSMakeRange(0, 64)), nil);
+			expect(requestedPaths()).to(equal(@[ @"/update.delta", @"/full.zip" ]));
+		});
+
+		it(@"should fall back to the full update when the delta fails part way through applying", ^{
```
MarshallOfSound, line 580: Nothing actually exercises the lchflags loop, this one leaves a partially written bundle behind but nothing in it is locked. If we `chflags uchg` one file in the running fixture for a variant of this (and one where the delta doesn't touch that file, per the thread above) it covers both the cleanup and the ShipIt case in one go.

MarshallOfSound (commented):
Re-swept 9bd1c83, everything from the first round is sorted (checked the submodule really is the 2.9.5 tag, and the digest memo gets cleared in the right place). One thing left over from the locked file thread plus a few small ones.

Squirrel/SQRLUpdater.m
codebytere: 298bceb clears file flags on the patched bundle before it is verified and staged (same loop, now `clearFileFlagsUnderURL:`, also run over a failed apply's leftovers before removal). New spec locks PkgInfo in the running fixture, which the delta leaves alone, and installs from the delta through ShipIt; it fails with the clearing knocked out.

Squirrel/SQRLUpdater.m
codebytere: Done: an apply or verify failure (not a download one) records the digest as unusable and later checks go straight to the zip. README updated to say so.

Squirrel/SQRLUpdater.m
codebytere: The fallback now removes its fresh directory itself when the zip side errors or has nothing new, and cleanUp skips a directory that is already gone, so no spurious log and nothing left for prune.

SquirrelTests/SQRLUpdaterSpec.m
codebytere: Added both: Info.plist locked (delta rewrites it: falls back, installs from the zip, and asserts nothing under update storage is still uchg) and PkgInfo locked (untouched: installs from the delta). Both fail with the two clearFileFlags calls removed and pass with them; 107 green.

MarshallOfSound (approved):

PR #331, opened 2026-09-06
fix: remove ShipIt's launchd job once installation completes

### LINKED ISSUE #21: Remove launchd job after update installation completes
It doesn't really hurt anything by being left there, but it might annoy advanced users who see a bunch of old jobs in their Library folder.

MarshallOfSound:
Closes #21.

Jobs submitted with `SMJobSubmit` are registrations, not one-shots, so after an update the ShipIt job lingers in the launchd domain as "not running" until logout. macOS 27 now surfaces this as a Dock tile on the updated app after the user quits it (see [this report](https://github.com/Squirrel/Squirrel.Mac/issues/21#issuecomment-5557307610)).

- ShipIt removes its own launchd job on the terminal success paths (install completed / cancelled because the app relaunched), just before `exit(0)`. Failure exits keep the job so `KeepAlive` still retries interrupted installs.
- Removal passes a NULL authorization: by then the installer has replaced the bundle containing the running ShipIt, so authd can't validate its on-disk signature and `AuthorizationCreate` fails with `errAuthorizationDenied` even as root. `SMJobRemove(NULL)` authorizes on the caller (root → system domain, otherwise the caller's user domain), never prompts, and works with the binary gone. This also means the 2013 concern above about re-prompting for the privileged path doesn't apply.
- `SIGTERM` is ignored during removal (launchd TERMs a job it's removing) and `wait=false` avoids deadlocking on our own exit.
- Test fixtures treat a vanished job as a clean exit, and a new spec asserts the job is gone after a remote install.

Validated beyond the suite (108/0): a system-domain end-to-end run of the built ShipIt as root — real 1.0→2.1 install with the privileged target guard satisfied — self-removed from the system domain with no authorization prompt, and deleted-binary probes confirm `SMJobRemove(NULL)` succeeds in both domains after the running binary is unlinked.

### DIFF
--- a/Squirrel/ShipIt-main.m
+++ b/Squirrel/ShipIt-main.m
@@ -17,6 +17,10 @@
 #include <sys/wait.h>
 #include <mach/mach.h>
 #include <servers/bootstrap.h>
+#include <signal.h>
+#include <unistd.h>
+
+#import <ServiceManagement/ServiceManagement.h>
 
 #import "NSError+SQRLVerbosityExtensions.h"
 #import "RACSignal+SQRLTransactionExtensions.h"
@@ -85,6 +89,50 @@ static void drainMachServicePort(const char *serviceName) {
 	                0, sizeof(msg), port, 0, MACH_PORT_NULL) == KERN_SUCCESS) {
 		mach_msg_destroy(&msg.header);
 	}
+}
+
+// Remove ShipIt's own launchd job on the way out. A job submitted with
+// SMJobSubmit is a registration, not a one-shot: after ShipIt exits, the
+// job would otherwise stay in the domain as "not running" until the login
+// session ends (or indefinitely, for the system domain), and macOS 27 shows
+// a Dock tile for any app with a registered background job — making an
+// updated-and-quit app look like it is still running. Removal is limited to
+// the terminal success paths; failure exits must keep the registration so
+// the KeepAlive/SuccessfulExit policy can respawn ShipIt to retry.
+//
+// SMJobRemove is deprecated but is the counterpart of the SMJobSubmit that
+// created the job (SQRLShipItLauncher); there is no other API that can
+// remove a submitted job.
+static void removeOwnLaunchdJob(NSString *jobLabel) {
+	// launchd terminates a running job as part of removing it. Ignore
+	// SIGTERM so we still exit through our own exit() call with the
+	// intended status rather than dying by signal mid-cleanup.
+	signal(SIGTERM, SIG_IGN);
+
+	// Privileged installs run ShipIt as root in the system domain;
+	// unprivileged ones run as the user in their domain.
+	//
+	// The authorization is deliberately NULL. By this point the installer
+	// has replaced (and typically deleted) the bundle containing this
+	// running ShipIt binary, so authd can no longer validate our code
+	// signature on disk and AuthorizationCreate fails with
+	// errAuthorizationDenied, even for root (verified empirically).
+	// SMJobRemove with a NULL authorization instead authorizes on the
+	// caller itself — root may modify the system domain and any caller its
+	// own user domain — which neither consults authd nor can present a
+	// prompt, and works with the binary already gone.
+	CFStringRef domain = (geteuid() == 0 ? kSMDomainSystemLaunchd : kSMDomainUserLaunchd);
+
+#pragma clang diagnostic push
+#pragma clang diagnostic ignored "-Wdeprecated-declarations"
+	// `wait` must be false: waiting blocks until the job (i.e. this
+	// process) exits, which would deadlock against our own exit().
+	CFErrorRef cfError = NULL;
+	if (!SMJobRemove(domain, (__bridge CFStringRef)jobLabel, NULL, false, &cfError)) {
+		NSError *error = CFBridgingRelease(cfError);
+		NSLog(@"Could not remove ShipIt launchd job %@: %@", jobLabel, error);
+	}
+#pragma clang diagnostic pop
 }
 
 // Waits for all instances of the target application (as described in the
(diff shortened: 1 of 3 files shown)


Squirrel/ShipIt-main.m
```diff
@@ -87,6 +91,50 @@ static void drainMachServicePort(const char *serviceName) {
...
+	CFStringRef domain = (geteuid() == 0 ? kSMDomainSystemLaunchd : kSMDomainUserLaunchd);
+
+#pragma clang diagnostic push
+#pragma clang diagnostic ignored "-Wdeprecated-declarations"
+	// `wait` must be false: waiting blocks until the job (i.e. this
+	// process) exits, which would deadlock against our own exit().
+	CFErrorRef cfError = NULL;
+	if (!SMJobRemove(domain, (__bridge CFStringRef)jobLabel, NULL, false, &cfError)) {
+		NSError *error = CFBridgingRelease(cfError);
+		NSLog(@"Could not remove ShipIt launchd job %@: %@", jobLabel, error);
```
VerteDinde, line 133: nit: worth swallowing `kSMErrorJobNotFound` here like `SQRLShipItLauncher` and the test fixture do — it happens legitimately if ShipIt is run outside launchd, or if a just-relaunched app's `launchPrivileged:` removes/resubmits the label before we get here.

Squirrel/ShipIt-main.m
```diff
@@ -87,6 +91,50 @@ static void drainMachServicePort(const char *serviceName) {
...
+	// The authorization is deliberately NULL. By this point the installer
+	// has replaced (and typically deleted) the bundle containing this
+	// running ShipIt binary, so authd can no longer validate our code
+	// signature on disk and AuthorizationCreate fails with
+	// errAuthorizationDenied, even for root (verified empirically).
+	// SMJobRemove with a NULL authorization instead authorizes on the
+	// caller itself — root may modify the system domain and any caller its
+	// own user domain — which neither consults authd nor can present a
+	// prompt, and works with the binary already gone.
+	CFStringRef domain = (geteuid() == 0 ? kSMDomainSystemLaunchd : kSMDomainUserLaunchd);
```
VerteDinde, line 124: q: this only clears the domain this run lives in. Someone who previously did an admin-prompted (system-domain) update keeps that stale job indefinitely and later unprivileged runs can't remove it, so the Dock tile could persist for them. Intentional / worth a note?

Squirrel/ShipIt-main.m
```diff
@@ -87,6 +91,50 @@ static void drainMachServicePort(const char *serviceName) {
...
+// the KeepAlive/SuccessfulExit policy can respawn ShipIt to retry.
+//
+// SMJobRemove is deprecated but is the counterpart of the SMJobSubmit that
+// created the job (SQRLShipItLauncher); there is no other API that can
+// remove a submitted job.
+static void removeOwnLaunchdJob(NSString *jobLabel) {
+	// launchd terminates a running job as part of removing it. Ignore
+	// SIGTERM so we still exit through our own exit() call with the
+	// intended status rather than dying by signal mid-cleanup.
+	signal(SIGTERM, SIG_IGN);
```
VerteDinde, line 110: q: if the install lands during logout/restart (the on-demand-only case) and the `SMJobRemove` IPC stalls, we now ignore launchd's TERM and hold the session for `ExitTimeOut` until KILL. Probably fine, but an `alarm()` here would bound it.

VerteDinde (commented):
Read through this against the launcher/updater side looking for ways the self-removal could bite (KeepAlive/Mach-demand respawn after `wait=false`, pgroup kill of the relaunched app, wrong domain, exit-0 paths that skip it, the app resubmitting concurrently) and it holds up. A few small things inline.

One heads-up for the Electron roll rather than this repo: Electron's GN `ShipIt` executable target only links AppKit/Foundation/IOKit/Security, so it'll need `ServiceManagement.framework` added when DEPS picks this up.

SquirrelTests/SQRLInstallerSpec.m
```diff
@@ -53,6 +54,23 @@ - (RACSignal *)deleteOwnedBundleAtURL:(NSURL *)bundleURL;
...
+
+	expect(self.testApplicationBundleVersion).to(equal(SQRLTestApplicationUpdatedShortVersionString));
+
+	// ShipIt removes its own job on success (SMJobRemove with wait=false),
+	// so the registration may lag its exit by a moment — poll for it to
+	// disappear rather than asserting immediately. A lingering registration
+	// causes macOS 27 to show a Dock tile for the updated app after quit.
+#pragma clang diagnostic push
+#pragma clang diagnostic ignored "-Wdeprecated-declarations"
+	expect(CFBridgingRelease(SMJobCopyDictionary(kSMDomainUserLaunchd, (__bridge CFStringRef)self.shipItDirectoryManager.applicationIdentifier))).withTimeout(SQRLLongTimeout).toEventually(beNil());
```
VerteDinde, line 70: nit: in the suite ShipIt runs from the test host's Squirrel.framework, so the running binary isn't inside the bundle being replaced. The production case (binary swapped for a differently-signed ShipIt at the same path, user domain, NULL auth) is only covered by the manual probes in the description — fine, just noting it isn't under CI.

SquirrelTests/QuickSpec+SQRLFixtures.m
```diff
@@ -338,7 +342,10 @@ - (void)submitShipItRequest:(SQRLShipItRequest *)request {
 #pragma clang diagnostic ignored "-Wdeprecated-declarations"
 static NSNumber *SQRLShipItLastExitStatus(NSString *jobLabel) {
 	NSDictionary *job = CFBridgingRelease(SMJobCopyDictionary(kSMDomainUserLaunchd, (__bridge CFStringRef)jobLabel));
-	if (job == nil || job[@"PID"] != nil) return nil;
+	// ShipIt removes its own launchd job when it finishes successfully, so
+	// once the job has been submitted, its disappearance means a clean exit.
+	if (job == nil) return @0;
```
VerteDinde, line 347: nit: this makes "never submitted" and "succeeded and removed" indistinguishable, so a launch failure now sails through `SQRLWaitForShipIt` and only trips on the later version assertion.

PR #332, opened 2026-09-07
build: replace the Xcode project with a GN build

codebytere:
Squirrel.Mac is now built and tested with GN and Ninja on top of Chromium's `//build`, which is how Electron has built it since 2020 using build files it carried downstream (originally written by @MarshallOfSound). This brings those files here, makes the repository build and test on its own with them, and removes the Xcode project and workspace, the git submodules and the objc-build-scripts. README.md has the new steps:

```sh
cp standalone.gclient .gclient && gclient sync
gn gen out/Default && ninja -C out/Default
script/test out/Default
```

`gclient sync` reads `DEPS`, which pins Chromium's `//build`, `//buildtools` and `tools/clang` plus the clang, GN and Ninja packages that go with them, and checks Mantle, ReactiveObjC and Sparkle (and Quick, Nimble and OHHTTPStubs for the tests) out under `vendor/` at the revisions the submodules were at. `.gn`, `build_overrides/build.gni` and a stub `third_party/llvm-libc/BUILD.gn` are what `//build` needs from an embedding project; the build uses the Xcode that `xcode-select` points at for the SDK, `swiftc` and `xctest` rather than Chromium's hermetic toolchain, keeps `//build`'s warning and SDK defaults, and targets macOS 11.0.

`BUILD.gn` and `filenames.gni` define `squirrel_framework` (with ShipIt in its Resources), `reactiveobjc_framework` and `mantle_framework` with the sources and flags the Xcode project used, and only use paths relative to themselves, so Electron consumes the same files at `//third_party/squirrel.mac` with the three libraries under its `vendor/`. Two differences from the files Electron carried: ReactiveObjC's public headers are staged under a generated `ReactiveObjC/` include directory for its own compile, since its sources import them framework-style and there is no Xcode header map, and the Sparkle delta sources keep default visibility so the tests can link Sparkle's delta generator against them. The test targets and `//:default` are only defined in a standalone checkout.

`SquirrelTests/BUILD.gn` builds the specs into `SquirrelTests.xctest` with the five frameworks under Contents/Frameworks and TestServer.py plus the signed TestApplication fixtures under Contents/Resources; `script/sign-test-fixtures` now runs as a build action on the built TestApplication.app, so the checked-in `TestApplication 2.1.app` and zip are gone. Quick and Nimble are mixed Swift/Objective-C and Chromium's macOS toolchain has no Swift support, so `gn/swift_module.gni` compiles their Swift halves with Xcode's `swiftc` and the toolchain links the object with the Objective-C halves; the specs include the generated `Quick-Swift.h`/`Nimble-Swift.h` through a prefix header because nothing is compiled with Clang modules. `SwiftSpec.swift`, which only checked that a Swift spec compiles, is dropped. TestApplication and its TestService XPC bundle get `BUILD.gn` files of their own; the app sets its delegate in `main()` instead of loading MainMenu.xib. `script/test` runs the bundle under `xctest` with the environment the scheme used to set.

CI runs the same steps on the existing macOS 14, 15 and 26 arm64 and x86_64 matrix, with depot_tools pinned to a revision in the workflow so every input is pinned. Nothing is cached: `gclient sync` takes under a minute and a restored `out/` would be older than the checkout, so Ninja would rebuild it anyway.


VerteDinde (commented):
Nice — verified parity where I could: same 107 tests as main on macOS 26 arm64, the target names Electron consumes are unchanged, vendor pins match Electron's DEPS, and the `//build` glue checks out against the pinned revision. A few inline notes, mostly about keeping upstream CI at least as strict as Electron's build.

Two non-inline things:
- Removing the xcodeproj ends Xcode/Carthage consumption of this repo. Probably nobody left, but worth tagging the last Xcode-buildable commit and saying so in the README/release notes.
- Electron follow-up on roll: `patches/squirrel.mac/build_add_gn_config.patch` and both `patches/ReactiveObjC/*` become redundant (the staged headers + `-Wno-implicit-const-int-float-conversion` replace them).

.gn
```diff
@@ -0,0 +1,24 @@
...
+script_executable = "python3"
+
+default_args = {
+  # Squirrel is Objective-C only; link the SDK's libc++ rather than building
+  # Chromium's, and skip Chromium's clang plugins and style checks.
+  use_custom_libcxx = false
+  use_custom_libcxx_for_host = false
+  use_llvm_libatomic = false
+  clang_use_chrome_plugins = false
+  treat_warnings_as_errors = false
```
VerteDinde: ci: `treat_warnings_as_errors = false` makes this CI laxer than Electron, which builds these same targets with the default `true` (`no_chromium_code` still adds `-Werror`). A new warning would pass here and then break the Electron roll. The build log is warning-free today, so I'd just drop this line.

.gn
```diff
@@ -0,0 +1,24 @@
...
+
+  # The oldest macOS the frameworks load on.
+  mac_deployment_target = "11.0"
+  mac_min_system_version = "11.0"
+
+  # The SDK in the oldest Xcode CI runs with.
+  mac_sdk_min = "14"
```
VerteDinde: nit: comment says "the SDK in the oldest Xcode CI runs with", but Xcode 16.2 ships the 15.2 SDK. Either call it a floor or bump to `"15"`.

build_overrides/build.gni
```diff
@@ -0,0 +1,15 @@
...
+import("//build/config/gclient_args.gni")
+
+enable_java_templates = false
+
+declare_args() {
+  # Build with the Xcode selected by `xcode-select` rather than Chromium's
+  # hermetic toolchain, which is not distributed outside Google.
+  use_system_xcode = true
+
+  ignore_elf32_limitations = false
```
VerteDinde: nit: nothing in the pinned `//build` or `//buildtools` reads `ignore_elf32_limitations`; looks inherited from an older template. Drop?

DEPS
```diff
@@ -0,0 +1,135 @@
...
+  'build_with_chromium',
+  'checkout_android',
+  'checkout_src_internal',
+]
+
+vars = {
+  'build_with_chromium': False,
+  'checkout_android': False,
+  'checkout_src_internal': False,
+  'checkout_clang_coverage_tools': False,
```
VerteDinde: nit: `checkout_clang_coverage_tools` is dead — not in `gclient_gn_args` and no dep condition uses it.

.github/workflows/ci.yml
```diff
@@ -28,85 +28,37 @@ jobs:
...
       - name: Select Xcode
         run: |
           sudo xcode-select -s "/Applications/Xcode_${{ matrix.xcode }}.app"
           xcodebuild -version
 
-      - name: Compute dep cache key
-        id: dep-key
+      - name: Install depot_tools
+        run: |
+          git clone --depth 1 https://chromium.googlesource.com/chromium/tools/depot_tools.git "$RUNNER_TEMP/depot_tools"
```
VerteDinde: nit: depot_tools HEAD is the one unpinned input here (actions are SHA-pinned, DEPS is pinned). `git clone` + `git checkout <rev>` for reproducibility?

BUILD.gn
```diff
@@ -0,0 +1,296 @@
...
+# Sparkle's BinaryDelta applier, configured by a force-included header in place
+# of Sparkle's xcconfig defines. Default visibility so the tests can link
+# Sparkle's delta generator against the archive classes in Squirrel.framework.
+source_set("sparkle_binary_delta") {
+  visibility = [ ":*" ]
+  sources = sparkle_binary_delta_filenames.sources
+  include_dirs = sparkle_binary_delta_filenames.include_dirs
+  configs -= [
+    "//build/config/compiler:chromium_code",
+    "//build/config/gcc:symbol_visibility_hidden",
```
VerteDinde, line 222: q: dropping `symbol_visibility_hidden` here means the shipping Squirrel.framework (Electron's included) now exports Sparkle's `SPU*`/bsdiff symbols, for a test-only consumer. Matches what the Xcode build did and two-level namespaces make it low-risk — just confirming that trade is intended.

SquirrelTests/BUILD.gn
```diff
@@ -0,0 +1,458 @@
+import("//build/config/mac/mac_sdk.gni")
+import("//build/config/mac/rules.gni")
+import("//gn/swift_module.gni")
+
+# XCTest and the Swift XCTest support libraries ship with the platform, not
+# the SDK. //build symlinks the platform and toolchain next to the SDK it hands
+# to the compiler, so derive their paths from that.
+_xcode_links = get_path_info(mac_sdk_path, "dir")
```
VerteDinde: nit: this derivation only holds when `mac_sdk_path` comes from `//build`'s `xcode_links` symlink farm; passing `mac_sdk_path=` explicitly gives the real `SDKs/` dir with no `MacOSX.platform` sibling. An assert or one-line comment would make that failure legible.

.gn
codebytere: dropped in 6ee5961 - the standalone build is back on the //build default, so as strict as Electron's.

.gn
codebytere: removed the override in 6ee5961; //build's own floor of 15 is met by the 15.2 SDK in Xcode 16.2, so there was nothing to relax.

build_overrides/build.gni
codebytere: yep, leftover from the pdfium template - gone in 6ee5961.

DEPS
codebytere: gone in 6ee5961.

.github/workflows/ci.yml
codebytere: pinned in 6ee5961: the workflow fetches `DEPOT_TOOLS_REVISION` (a sha at the top of the file, with a note on when to bump it) and `DEPOT_TOOLS_UPDATE=0` keeps it there.

BUILD.gn
codebytere: intended, yeah. the tests link Sparkle's delta generator against the archive classes inside Squirrel.framework so there is one copy of them in the process, which is also why `SparkleDeltaConfig.h` turns off `SPU_OBJC_DIRECT`, and the Xcode build exported them the same way. compiling a second copy into the test bundle would mean duplicate ObjC class definitions at runtime, which i'd rather avoid, so leaving this as is.

SquirrelTests/BUILD.gn
codebytere: reworked in 6ee5961: `SquirrelTests/BUILD.gn` now asks //build's `sdk_info.py` for the platform and toolchain paths (allowlisted in `.gn`) instead of assuming the xcode_links layout, so an explicit `mac_sdk_path` works too.

VerteDinde (approved):

MarshallOfSound (approved):

PR #259, opened 2021-06-18
align terminationHandler with dealloc in SQRLZipArchiver

MarshallOfSound:
For some reason even though the `NSTask` is terminated the file handle is still being read.  Fixes #247

Verified locally using the electron test suite and `chown nobody:nogroup Electron.app` during the test suite

Issue #247, opened 2020-04-27
Squirrel-based Install Dialogue Burns CPU/Fans on Mac


ambroselittle:
Please [see this bug filed with VS Code for context/details](https://github.com/microsoft/vscode/issues/95534). They suggested I file this with you since they are using this framework.

Let me know if you have any questions!

ambroselittle:
Please note we found a workaround on the linked issue. Not sure if that's something that can be fixed here or not.

stk-ableton:
This is a very annoying and long-standing problem, and it would be great to get it fixed. The workaround that @ambroselittle refers to in the comment above is not a real solution, because it involves chown-ing the application to the current user, which is not something that you want to do from a security point of view.

The problem occurs whenever the dialog "An update is ready to install." is on the screen, asking for an admin password; as long as this dialog is open, there's a very high CPU usage. I've seen this happen consistently in all apps that use Squirrel, e.g. WhatsApp, Signal, VS Code, and others. It is a problem because many times these apps start the update process in the night, so when I come to the machine in the morning, it's been running with fans on full speed for hours already.

And it's particularly problematic when you use the VS Code Insiders version, because then there's an update every day (or night).

So I tried to investigate a little where this comes from. Running Instruments seems to indicate that it comes from https://github.com/Squirrel/Squirrel.Mac/blob/master/Squirrel/SQRLZipArchiver.m#L70, which seems to be called repeatedly while the dialog is open.

I don't understand enough about Cocoa and ReactiveCocoa to say what's the best fix for this; one wild guess is that you should set `standardErrorPipe.fileHandleForReading.readabilityHandler` back to `nil` when the ditto task completes. Another wild guess is that there's something wrong with the lifetime of the `SQRLZipArchiver` object; it seems to be still around after `unarchiveAndPrepareData` returns, but maybe shouldn't, and it looks like that would also fix the issue.

stk-ableton:
Just a friendly ping. Anybody?

Cc:ing @jspahrsummers who touched this code last.

jspahrsummers:
Very sorry, but I haven't worked on this codebase in ~6 years. I'm not knowledgeable about it anymore.

tmoschou:
[...]

PR #333, opened 2026-09-12
fix: stop the stderr readability handler at EOF and on task termination

alexkozy:
## Summary

`SQRLZipArchiver` installs an `NSFileHandle.readabilityHandler` on `ditto`'s stderr pipe and never removes it. Once the pipe reaches EOF, Foundation keeps invoking the handler with empty data until the handler is set to `nil`; each invocation is forwarded into the `collect` behind `standardErrorData`, so the array grows without bound until the host process dies. This change removes the handler as soon as `availableData` comes back empty, and detaches it before `closeFile` in both the termination handler and `dealloc`.

## Crash

Observed in a production Electron app on stock Electron 40.10.3 and 42.10.0 (both pin Squirrel.Mac `0e5d146`), on macOS 15.5 through 26.x. Symbolicated browser-process crash dumps all end in the same place:

```
partition_alloc::internal::PartitionExcessiveAllocationSize   (single allocation over the ~2 GiB direct-map limit)
CoreFoundation                                                (NSMutableArray growth)
-[RACSignal(Operations) collect]_block_invoke_2               RACSignal+Operations.m:369
-[RACSubject sendNext:]
-[SQRLZipArchiver init]_block_invoke                          Squirrel/SQRLZipArchiver.m:71
Foundation / libdispatch                                      (readability source)
```

Process telemetry for the same crashes shows the main process at roughly two cores with resident memory growing linearly for about eleven minutes after `ditto` exits, on machines whose update pipeline had otherwise finished. That is the signature of a handler spinning on an EOF pipe, and `closeFile` alone (added in #259 for the CPU-burn variant of the same loop, #247) is not enough to stop it.

## Change

- `SQRLZipArchiver.m`: the readability handler sets itself to `nil` and returns when `availableData` is empty; the termination handler and `dealloc` set it to `nil` before closing the file handle, which is the order Foundation documents for tearing down a handler.
- `SQRLZipArchiverSpec.m`: a spec runs a failing `/bin/sh` task through `launchWithArguments:` and asserts the handler is `nil` once the launch signal has completed.

Behaviour on the success and failure paths is unchanged: exit status and the stderr text collected before termination still flow into the error's `NSLocalizedDescription`.

## Relation to #324

#324 addresses the same crash and additionally caps the retained stderr at 1 MiB. This PR is the minimal form: only the handler lifecycle. If #324 lands first, this can be closed.

## Testing

The new spec runs under `script/test` in CI. I do not have a macOS machine available to run it locally.

### DIFF
--- a/Squirrel/SQRLZipArchiver.m
+++ b/Squirrel/SQRLZipArchiver.m
@@ -62,13 +62,21 @@ const NSInteger SQRLZipArchiverShellTaskFailed = 1;
 		@strongify(self);
 		if (self == nil) return;
 
+		NSFileHandle *standardError = self.standardErrorPipe.fileHandleForReading;
+		standardError.readabilityHandler = nil;
 		[self->_taskTerminated sendNext:@(task.terminationStatus)];
-		[self.standardErrorPipe.fileHandleForReading closeFile];
+		[standardError closeFile];
 	};
 
 	RACSubject *errorDataChunks = [[RACSubject subject] setNameWithFormat:@"errorDataChunks"];
 	self.standardErrorPipe.fileHandleForReading.readabilityHandler = ^(NSFileHandle *handle) {
-		[errorDataChunks sendNext:handle.availableData];
+		NSData *data = handle.availableData;
+		if (data.length == 0) {
+			handle.readabilityHandler = nil;
+			return;
+		}
+
+		[errorDataChunks sendNext:data];
 	};
 
 	_standardErrorData = [[[[[[errorDataChunks
@@ -91,7 +99,9 @@ const NSInteger SQRLZipArchiverShellTaskFailed = 1;
 
 - (void)dealloc {
 	[_taskTerminated sendCompleted];
-	[self.standardErrorPipe.fileHandleForReading closeFile];
+	NSFileHandle *standardError = self.standardErrorPipe.fileHandleForReading;
+	standardError.readabilityHandler = nil;
+	[standardError closeFile];
 }
 
 #pragma mark Archiving/Unarchiving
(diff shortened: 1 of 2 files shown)


SquirrelTests/SQRLZipArchiverSpec.m
```diff
@@ -51,6 +52,18 @@ - (RACSignal *)launchWithArguments:(NSArray *)arguments;
...
 
+it(@"should stop reading standard error once the ditto task has terminated", ^{
+	SQRLZipArchiver *archiver = [[SQRLZipArchiver alloc] init];
+	archiver.dittoTask.launchPath = @"/bin/sh";
+
+	NSError *error = nil;
+	BOOL success = [[archiver launchWithArguments:@[ @"-c", @"echo failure >&2; exit 1" ]] asynchronouslyWaitUntilCompleted:&error];
+
+	expect(@(success)).to(beFalsy());
+	expect(@(error.code)).to(equal(@(SQRLZipArchiverShellTaskFailed)));
```
codebytere, line 63: could you also assert `error.domain` here like the spec above does? `RACSignalErrorTimedOut` is also code 1, so as written a hang that times out in `asynchronouslyWaitUntilCompleted:` would pass.

Squirrel/SQRLZipArchiver.m
```diff
@@ -62,13 +62,21 @@ - (id)init {
 		@strongify(self);
 		if (self == nil) return;
 
+		NSFileHandle *standardError = self.standardErrorPipe.fileHandleForReading;
+		standardError.readabilityHandler = nil;
 		[self->_taskTerminated sendNext:@(task.terminationStatus)];
-		[self.standardErrorPipe.fileHandleForReading closeFile];
+		[standardError closeFile];
```
codebytere, line 68: now that the handler cancels itself at EOF i think we can drop `closeFile` here and in `-dealloc` entirely - `NSPipe` closes the descriptor when it's released, and closing it from the termination thread can still race the one readability invocation that's already in flight (EBADF from `availableData`, or the fd number getting reused). setting the handler to nil in both places is enough.

codebytere (commented):
thanks for this, the EOF self-nil is the right fix. You'll also need to fix CI. two small things below and then this is good to go imo.

